PR #22, opened 2015-09-16
remove - on julia version in REQUIRE

tkelman:
that refers to prereleases of julia 0.3 and has not been necessary for over a year


spencerlyon2:
Thanks.

PR #23, opened 2015-10-07
Fix type for singular problems

timholy:
Without the change to trust_region.jl, the new test would trigger:
```jl
ERROR: MethodError: `wnorm` has no method matching wnorm(::Array{Float32,1}, ::Array{Float64,1})
```


spencerlyon2:
Looks great. Thanks

PR #24, opened 2015-12-28
fix bug in objective function creator in newton.jl

KristofferC:
The previous version simply bounded the input argument to an array but this doesn't do what is desired which is to update the actual input storage vector.

Without this I have problems where my iterations stalls with something like:

```jl
Iter     f(x) inf-norm    Step 2-norm 
------   --------------   --------------
     0     1.554002e+07              NaN
     1     4.528678e-01     1.586823e+14
     2     4.528678e-01     0.000000e+00
     3     4.528678e-01     0.000000e+00
     4     4.528678e-01     0.000000e+00
     5     4.528678e-01     0.000000e+00
     6     4.528678e-01     0.000000e+00
     7     4.528678e-01     0.000000e+00
     8     4.528678e-01     0.000000e+00
```

with this patch it works.

cc @timholy @sebastien-villemot @spencerlyon2

### DIFF
--- a/src/newton.jl
+++ b/src/newton.jl
@@ -28,11 +28,11 @@ function create_objective_function(df::AbstractDifferentiableMultivariateFunctio
     end
     function go!(x::Vector{T}, storage::Vector{T})
         df.fg!(x, fvec2, fjac2)
-        storage = fjac2'*fvec2
+        copy!(storage, fjac2'*fvec2)
     end
     function fgo!(x::Vector{T}, storage::Vector{T})
         df.fg!(x, fvec2, fjac2)
-        storage = fjac2'*fvec2
+        copy!(storage, fjac2'*fvec2)
         return(dot(fvec2, fvec2)/2)
     end
 


spencerlyon2:
Great, thanks! when I see the green light from travis I'll merge

KristofferC:
I have some thoughts of integrating https://github.com/JuliaDiff/ForwardDiff.jl with this package. Right now the autodiff functionality is just using dual numbers and have to iterate to get the sensitivity for all components which is slow.

Would that be a wanted patch?

spencerlyon2:
I think the autodiff functionality currently here was added long before ForwardDiff is what it is today.

I would like to see that patch. 

Do you anticipate performance changes when moving autodiff backends?

KristofferC:
Performance should be significantly improved since jacobians for example are computed in one function call in ForwardDiff while with using one component dual numbers you need `n` calls.

I could do some small benchmarks after I've implemented it.

spencerlyon2:
Sounds great, thanks!

KristofferC:
Maybe it is worth to tag a new version now when this bug fix is in?

cc @spencerlyon2

spencerlyon2:
yep, will do. Thanks for the reminder

PR #25, opened 2015-12-28
Modernize test suite

KristofferC:
This modernizes the test suite to use the `@test` macros together with using the https://github.com/IainNZ/BaseTestNext.jl package which is the new test merged into v0.5.

I pushed this on top of #24 since I expect that to get merged and wanted to save a rebase.


spencerlyon2:
Thanks!

PR #26, opened 2015-12-28
change autodiff backend to ForwardDiff

KristofferC:
TODO before merge

- [x] Wait until a new version of ForwardDiff is tagged. There are some bugs that are fixed on master in ForwardDiff but needs a tag,

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Alternatively, you can construct an object of type
 ### Automatic differentiation
 
 Another option if you do not have a function computing the Jacobian is to use
-automatic differentiation, thanks to the `DualNumbers` package. The syntax is
+automatic differentiation, thanks to the `ForwardDiff` package. The syntax is
 simply:
 
     nlsolve(f!, initial_x, autodiff = true)
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.3
 Optim
 Calculus
-DualNumbers
+ForwardDiff
 Compat
 Distances
--- a/src/NLsolve.jl
+++ b/src/NLsolve.jl
@@ -4,6 +4,7 @@ module NLsolve
 
 using Distances
 using Optim
+using ForwardDiff
 
 import Base.show,
        Base.push!,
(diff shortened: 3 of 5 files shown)


KristofferC:
Pkg.checkout seems to resolve the packages so the `Distances` I moved to the test REQUIRE file gets removed. Anyway, I am tagging a new version of ForwardDiff now, will update this when it gets merged into METADATA.

KristofferC:
ForwardDiff tagged and commits squashed and passing.

Need to get some good stuff to benchmark. Although I am pretty sure @jrevels and @mlubin has already done significant benchmarks comparisons between `DualNumbers` and `ForwardDiff` so maybe they can chime in.

mlubin:
The default vector mode could be slower than ``DualNumbers``, you should use chunk mode but that's not fully general unless you set ``chunk_size=1``.

jrevels:
> you should use chunk mode but that's not fully general unless you set chunk_size=1.

Details on this can be found in JuliaDiff/ForwardDiff.jl#36, if you're curious

KristofferC:
Ok so I added an extra optional argument to `nlsolve` when using `:autodiff = true` which is the `chunk_size` that is defaulted to 1. I also added some small note in the README about `chunk_size` and a link to the ForwardDiff documentation about how to chose it.

mlubin:
I don't think ``chunk_size`` should really be a user-facing option at this level. As soon as JuliaDiff/ForwardDiff.jl#36 is resolved you should be able to set ``chunk_size=10`` by default and forget about it.

KristofferC:
So like this then by just setting `chunk_size` to 1 with a comment about https://github.com/JuliaDiff/ForwardDiff.jl/issues/36.

Would performance for this be on par with `DualNumbers`?

mlubin:
Yes, performance should be about the same as ``DualNumbers`` in that case and the API is simpler, so still an overall win.

spencerlyon2:
Thanks

KristofferC:
Thanks for your valuable inputs @mlubin and @jrevels!

And thanks @spencerlyon2 for merge.

PR #29, opened 2015-12-31
remove unused call to dfo!

KristofferC:
AFAIU this call does not do anything.

The return value is not used and the gr-vector is simply passed to the linesearch where it is immediately overwritten, https://github.com/JuliaOpt/Optim.jl/blob/d9f959480b8ced00572f3140dd85822e74fc0d03/src/linesearch/backtracking_linesearch.jl#L26


KristofferC:
Subsumed by #33

PR #30, opened 2015-12-31
Split up NLSolve.jl and introduce IsFiniteExcpetion

KristofferC:
First commit fixes the bloat in the file `NLsolve.jl`. There were many type definitions and unrelated function definitions there so I separated them out in a few files

Second commit introduces an "IsFiniteException" that is thrown when a non finite value is encountered in the function value. Before, this was performedn in a `error("...")` block but using proper exceptions is more idiomatic. I also included tests for the exceptions.


src/newton.jl
```diff
@@ -96,8 +92,6 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
         df.g!(x, fjac)
         g_calls += 1
 
-        g = fjac'*fvec
```
KristofferC: Forgot that I added this. g is not used except in the catch block so move it in there to save a matvec multiplication

spencerlyon2:
Hey  @KristofferC with the other PRs that have been merged, what is the status of this PR?

KristofferC:
I just need to rebase it.

KristofferC:
Rebased.

spencerlyon2:
Thanks! I think we are ready for a METADATA re-tag now

KristofferC:
Yes, after #34 is merged I have no immediate plans on adding or changing anything so a re-tag would be great.

PR #33, opened 2016-01-01
fix not double computing g! in each newton iteration

### LINKED ISSUE #32: Gradient called twice per iterations.
As it is right now the gradient function is called twice per iteration in the newton scheme.

The first one here: https://github.com/EconForge/NLsolve.jl/blob/ae1e89a5795324a7a35f82a22499932cc1f569f7/src/newton.jl#L96

and the second one in the linesearch i.e: https://github.com/JuliaOpt/Optim.jl/blob/master/src/linesearch/backtracking_linesearch.jl#L26

This is a bit unfortunate since the gradient is sometimes quite expensive to compute.

We could fix this by "cheating" by knowing that the linesearches only computes the gradients at the current point where we have already computed the derivative.

As an example, consider this test case with an artificial sleep introduced in the gradient:

```jl
function f!(x, fvec)
    fvec[1] = (x[1]+3)*(x[2]^3-7)+18
    fvec[2] = sin(x[2]*exp(x[1])-1)
end

function g!(x, fjac)
    fjac[1, 1] = x[2]^3-7
    fjac[1, 2] = 3*x[2]^2*(x[1]+3)
    u = exp(x[1])*cos(x[2]*exp(x[1])-1)
    fjac[2, 1] = x[2]*u
    fjac[2, 2] = u
    sleep(1)
end

df = DifferentiableMultivariateFunction(f!, g!)
```

I get this on master (7 seconds):
```jl
julia> @time r = nlsolve(df, [ -0.5; 1.4], method = :newton, linesearch! = Optim.backtracking_linesearch!, ftol = 1e-6)
  7.020818 seconds (237 allocations: 10.500 KB)
Results of Nonlinear Solver Algorithm
 * Algorithm: Newton with line-search
 * Starting Point: [-0.5,1.4]
 * Zero: [-1.779077697780431e-8,1.0000000055513854]
 * Inf-norm of residuals: 0.000000
 * Iterations: 3
 * Convergence: true
   * |x - x'| < 0.0e+00: false
   * |f(x)| < 1.0e-06: true
 * Function Calls (f): 10
 * Jacobian Calls (df/dx): 6
end
```

Using #33 I get (3 seconds):

```jl
julia> @time r = nlsolve(df, [ -0.5; 1.4], method = :newton, linesearch! = Optim.backtracking_linesearch!, ftol = 1e-6)
  3.009181 seconds (189 allocations: 8.359 KB)
Results of Nonlinear Solver Algorithm
 * Algorithm: Newton with line-search
 * Starting Point: [-0.5,1.4]
 * Zero: [-1.779077697780431e-8,1.0000000055513854]
 * Inf-norm of residuals: 0.000000
 * Iterations: 3
 * Convergence: true
   * |x - x'| < 0.0e+00: false
   * |f(x)| < 1.0e-06: true
 * Function Calls (f): 10
 * Jacobian Calls (df/dx): 3
end
```

It is a bit ugly and we have to double check that it works with the other line searches but it is my opinion that we need to do something to fix this because double calling the graident is unacceptable.

KristofferC:
This fixes #32 by checking if the gradient queried in the line search is the same as the one we just computed in newton iterations. Effectively halves the number of gradient queries at the cost of one vector equality check. See #32 for a runnable example.

### DIFF
--- a/src/newton.jl
+++ b/src/newton.jl
@@ -18,28 +18,6 @@ macro newtontrace(stepnorm)
     end
 end
 
-function create_objective_function(df::AbstractDifferentiableMultivariateFunction,
-                                   T::Type, nn::Integer)
-    fvec2 = Array(T, nn)
-    fjac2 = alloc_jacobian(df, T, nn)
-    function fo(x::Vector{T})
-        df.f!(x, fvec2)
-        return(dot(fvec2, fvec2)/2)
-    end
-    function go!(x::Vector{T}, storage::Vector{T})
-        df.fg!(x, fvec2, fjac2)
-        copy!(storage, fjac2'*fvec2)
-    end
-    function fgo!(x::Vector{T}, storage::Vector{T})
-        df.fg!(x, fvec2, fjac2)
-        copy!(storage, fjac2'*fvec2)
-        return(dot(fvec2, fvec2)/2)
-    end
-
-    return(DifferentiableFunction(fo, go!, fgo!))
-end
-
-
 function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
                    initial_x::Vector{T},
                    xtol::T,
@@ -55,6 +33,7 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
     xold = fill(convert(T, NaN), nn)
     fvec = Array(T, nn)
     fjac = alloc_jacobian(df, T, nn)
+
     p = Array(T, nn)
     g = Array(T, nn)
     gr = Array(T, nn)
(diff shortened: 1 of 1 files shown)


src/newton.jl
```diff
@@ -84,10 +63,38 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
...
+            df.g!(xlin, fjac2)
+            copy!(storage, fjac2'*fvec2)
+        end
+        # For convenience, return fo(x) here so that fgo(x) = go(x)
+        return(dot(fvec2, fvec2)/2)
+    end
+    fgo!(xlin::Vector{T}, storage::Vector{T}) = go!(xlin, storage)
 
-    # Keep track of step-sizes
-    foval = dfo.fg!(x, gr)
```
KristofferC: See #29

PR #34, opened 2016-01-04
Fix missing space in exception message

KristofferC:
Sorry, just found this @spencerlyon2


spencerlyon2:
No problem, thanks for all the work here. Merging before CI as this is a whitespace change in a string

PR #35, opened 2016-01-29
Add DifferentiableGivenSparseMultivariateFunction

matthieugomez:
When solving a problem with sparse Jacobian, it's sometimes faster to write the `g!` function assuming that the given jacobian matrix has a certain sparsity structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,14 @@ can be achieved with:
     empty!(gx.rowval)
     empty!(gx.nzval)
 
+
+Another solution is to directly pass a Jacobian matrix with a given sparsity. To do so, construct an object of type `DifferentiableGivenSparseMultivariateFunction`
+
+    df = DifferentiableGivenSparseMultivariateFunction(f!, g!, J)
+    nlsolve(df, initial_x)
+
+If  `g!` conserves the sparsity structure of `gx`, `gx` will always have the same sparsity as `J`. This sometimes allow to write a faster version of `g!`.
+
 # Fine tunings
 
 Two algorithms are currently available. The choice between the two is achieved
--- a/src/differentiable_functions.jl
+++ b/src/differentiable_functions.jl
@@ -113,3 +113,22 @@ function DifferentiableSparseMultivariateFunction(f!::Function, g!::Function)
     end
     return DifferentiableSparseMultivariateFunction(f!, g!, fg!)
 end
+
+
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -200,6 +200,14 @@ can be achieved with:
     empty!(gx.rowval)
     empty!(gx.nzval)
 
+
+Another solution is to directly pass a Jacobian matrix with a given sparsity. To do so, construct an object of type `DifferentiableGivenSparseMultivariateFunction`
+
+    df = DifferentiableGivenSparseMultivariateFunction(f!, g!, J)
+    nlsolve(df, initial_x)
+
+If  `g!` conserves the sparsity structure of `gx`, `gx` will always have the same sparsity than `J`. This sometimes allow to write a faster version of `g!`.
```
KristofferC: > same sparsity **as** `J`?

KristofferC:
Could you rebase on master now when I fixed it on 0.5?

matthieugomez:
I'm not sure I did it but I tried!

KristofferC:
Heh, I renamed some test functions because we got warnings about function redefinitions. You need to update your test.

KristofferC:
LGTM 

@spencerlyon2?

spencerlyon2:
Sorry for the delay. Yep looks good, thanks.

I will say I don't love the names, but you've totally followed suit with what was here so that isn't a problem with this PR.

src/differentiable_functions.jl
```diff
@@ -113,3 +113,22 @@ function DifferentiableSparseMultivariateFunction(f!::Function, g!::Function)
...
+
+
+immutable DifferentiableGivenSparseMultivariateFunction{Tv, Ti} <: AbstractDifferentiableMultivariateFunction
+    f!::Function
+    g!::Function
+    fg!::Function
+    J::SparseMatrixCSC{Tv, Ti}
+end
+
+alloc_jacobian(df::DifferentiableGivenSparseMultivariateFunction, args...) = deepcopy(df.J)
```
KristofferC: Why `deepcopy` here?

matthieugomez: Otherwise, `df.J` will be modified within the algorithm, right? Would not that be error-prone?

KristofferC: Shouldn't a normal copy work? 

Whether it is safe or not to modify the input matrix or not is debatable. Personally I wouldn't mind it being modified because the reason I created the sparse matrix in the first place is to use it in the solver.

PR #37, opened 2016-02-03
use identically equal in comparing input vector to linesearch gradient

KristofferC:
The linesearches in Optim does not copy the input vectors so there is no point in comparing element by element. Instead we can just compare using ===.


KristofferC:
Actaully this doesnt work.

PR #39, opened 2016-02-04
improve autodiff option

KristofferC:
- Also generate a fg! function using the `ForwardDiff.AllResult` option, see http://www.juliadiff.org/ForwardDiff.jl/lower_order_results.html
- Try to find a chunk_size that divides the input length but is smaller or equal to `ForwardDiff.tuple_usage_threshold`

Does this look reasonable @jrevels @mlubin? Thanks

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.4
 Optim
 Calculus
-ForwardDiff
+ForwardDiff 0.1.4 0.2
 Distances
--- a/src/autodiff.jl
+++ b/src/autodiff.jl
@@ -2,12 +2,39 @@
 # assuming that f takes a Vector{T} of length n
 # and writes the result to a Vector{T} of length m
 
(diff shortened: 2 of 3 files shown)


mlubin:
I wouldn't depend on ``ForwardDiff.tuple_usage_threshold`` because that will likely disappear soon, right @jrevels?

KristofferC:
Yes, but I specified the specific ForwardDiff release because of that. I will update this pretty fast after a new ForwardDiff is tagged anyway.

mlubin:
It's safer to put:
```
ForwardDiff 0.1.4 0.2
```

jrevels:
It'll disappear soon, but those changes will only initially target Julia v0.5 (there's a lot of backportable stuff there, however).

spencerlyon2:
LGTM

KristofferC:
I will merge this then and deal with updates to ForwardDiff when they come :)

PR #41, opened 2016-02-04
add test for counting number of function and gradient calls 

KristofferC:
This adds a test file to make sure that we are counting the number of function and gradient calls correctly.

This is currently true but I am experimenting with some changes and don't want to break anything.

It also creates a file with the number of function calls + gradient calls + residual to check for regressions. It does not track computational time since that is out of scope and would vary between systems.

### DIFF
--- a/test/f_g_counts.jl
+++ b/test/f_g_counts.jl
@@ -1,0 +1,40 @@
+# This file tests if the number of f and g calls are correctly counted
+@testset "f_g_counts" begin
+
+    fcalls = Ref(0)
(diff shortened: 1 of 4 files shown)


KristofferC:
It is also fun to see the numbers go down when you are making changes :)

spencerlyon2:
Cool, thanks

KristofferC:
Hmm, I noticed that this nly tests the function with trust_region, will see if they pass with Newton...

PR #43, opened 2016-02-04
improvements to reduce number of newton function and gradient calls

KristofferC:
This is the point of my previous PRs.

It is easy to see how the number of function calls are reduced while keeping the same error.


spencerlyon2:
:100: good work

PR #44, opened 2016-02-04
remove unnecessary call to f!

### LINKED ISSUE #40: Double computing function value in Newton iterations due to backtracking
Not all linesearches in `Optim` returns the computed function value. Instead we unnecessarily have to recompute it, even though we just evaluated the function at the same point. https://github.com/EconForge/NLsolve.jl/blob/ff78c8e1323c0038db8a95a20ce8250bf3b9f86b/src/newton.jl#L130

We should probably look into getting the backtracking functions in Optim return the function value and we can then skip a function evaluation for each newton iteration.

KristofferC:
This is pretty funny...

Fixes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,12 @@ In order to find a zero of this function and display it, you would write the
 following program:
 
     using NLsolve
-    
+
     function f!(x, fvec)
         fvec[1] = (x[1]+3)*(x[2]^3-7)+18
         fvec[2] = sin(x[2]*exp(x[1])-1)
     end
-    
+
     function g!(x, fjac)
         fjac[1, 1] = x[2]^3-7
         fjac[1, 2] = 3*x[2]^2*(x[1]+3)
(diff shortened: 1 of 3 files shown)


KristofferC:
What we need to think about here is that we assume that the linesearch will actually evaluate the function. A user who passes his own line search along the lines of `linesearch!(x...) = 0.0, 0, 0` will be out of luck.

We should probably think about this and add it to the documentation.

KristofferC:
Hmm, forgot that you can't overload keyword argument with type dispatch.. have to think of something else.

KristofferC:
For now, I think it is fine if we document it and then tag a new 0.7 version. If anyone screams we can deal with it.

spencerlyon2:
Sounds good to me.




// Spencer

On February 4, 2016 at 1:09:04 PM, Kristoffer Carlsson (notifications@github.com) wrote:

For now, I think it is fine if we document it and then tag a new 0.7 version. If anyone screams we can deal with it.

PR #51, opened 2016-05-16
RFC: run without linesearch by default in newton solver

KristofferC:
What do you think @spencerlyon2 

The backwards linesearch in Optim is almost always slower than the full newton for the problems I have tested.


spencerlyon2:
I'm happy with this change. 

What makes the line search slower?

KristofferC:
Not sure why, maybe it is just on my type of problems or that the Optim function has a lot of overhead. No matter, it feels sensible to try a normal newton by default and if the user wants to enable a linesearch, it is very easy to enable.

KristofferC:
If you are happy, then I am happy.

spencerlyon2:
👍 looks good to me!

PR #54, opened 2016-06-19
upgrade to new ForwardDiff

KristofferC:
Also extend the newton definition in Optim so we dont clash with exported functions.

Does the AD stuff look ok to you @jrevels? I wasn't able to get rid of the `fx2 = copy(initial_x)` copy because the function that computes the gradient in place does not have access to the output buffer for the function and this is now needed in ForwardDiff if you want to use the old `mutates=true`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.4
 Optim
 Calculus
-ForwardDiff 0.1.4 0.2
 Distances
-Compat 0.7.16
+ForwardDiff 0.2.0
+Compat 0.8.0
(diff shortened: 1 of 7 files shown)


KristofferC:
However, previously, ForwardDiff allocated that array by itself so it is not like it matters.

jrevels:
LGTM! Seeing this diff makes me so happy that I got rid of the function generating API methods, it's so much cleaner now.

KristofferC:
Could have squashed :P

spencerlyon2:
Ah yes, sorry. I'm still getting used to that.

Thanks for the patch!

PR #55, opened 2016-06-25
Add reference to JuMP

matthieugomez:
Helpful tip when NLsolve does not converge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ julia> fvec
 
 # Related Packages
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -369,6 +369,10 @@ julia> fvec
 * Homotopy methods
 * [LMMCP algorithm by C. Kanzow](http://www.mathematik.uni-wuerzburg.de/~kanzow/)
 
+# Related Packages
+
+* JuMP.jl can also solve non linear equations. To reformulate your problem as an optimization problem with non linear constraint, use the set of equations as constraints, and enter 1.0 as the objective function. JuMP currently supports a number of open-source and commercial solvers.
```
KristofferC: non linear constraint**s**?

Probably worth to make the Jump.jl into a hyperlink.

PR #63, opened 2016-08-30
in place inversion for dense matrices

matthieugomez:


KristofferC:
I remember experimenting with this a bit but found it to be slower somehow.. Do you have a benchmark? Preferably one with smallish `n` and biggish `n`.

matthieugomez:
You're right.

PR #65, opened 2016-09-04
Add broyden method

matthieugomez:
Solve #1

### DIFF
--- a/src/broyden.jl
+++ b/src/broyden.jl
@@ -87,9 +87,9 @@ function broyden_{T}(df::AbstractDifferentiableMultivariateFunction,
             catch e
                 if isa(e, Base.LinAlg.LAPACKException)
                     fjac2 = Ac_mul_B(fjac, fjac)
-                    lambda = convert(T,1e6)*sqrt(nn*eps())*norm(fjac2, 1)
+                    lambda = convert(T, 1e6) * sqrt(nn * eps()) * norm(fjac2, 1)
                     g = Ac_mul_B(fjac, fvec)
-                    p = (fjac2 + lambda*eye(nn))\g
+                    p = (fjac2 + lambda * eye(nn)) \ g
                     scale!(p, -1)
                 else
                     throw(e)
@@ -106,7 +106,7 @@ function broyden_{T}(df::AbstractDifferentiableMultivariateFunction,
             d = dot(p, minvjΔf)
             Base.axpy!(1.0, p, minvjΔf)
             scale!(minvjΔf, - 1 / d)
-            A_mul_Bt!(scaleinvfjac, minvjΔf , p)
+            A_mul_Bc!(scaleinvfjac, minvjΔf , p)
             for i in 1:size(scaleinvfjac, 2)
                 scaleinvfjac[i, i] += 1.0
             end


KristofferC:
cc @jmartin3 @stevengj, If anyone of you have time you are probably better to review this than me.

sglyon:
I'd like to try to merge this before we move orgs... just so things start out in a cleaner state. 

I can try to find time to review, but if anyone else is interested and has time that'd be great.

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
+# based on 
+# A QUASI-NEWTON METHOD FOR SOLVING SMALL NONLINEAR SYSTEMS OF ALGEBRAIC EQUATIONS
+# JEFFREY MARTIN
+# http://web.mit.edu/jmartin3/Public/Project.pdf
+macro broydentrace(stepnorm)
```
stevengj: Why is this a macro?

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
...
+    quote
+        if tracing
+            dt = Dict()
+            if extended_trace
+                dt["x"] = copy(x)
+                dt["f(x)"] = copy(fvec)
+                dt["g(x)"] = copy(fjac)
```
stevengj: Seems weird to use a dictionary for this.

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
...
+            try
+                p = fjac \ fvec
+                scale!(p, -1)
+            catch e
+                if isa(e, Base.LinAlg.LAPACKException)
+                    fjac2 = Ac_mul_B(fjac, fjac)
+                    lambda = convert(T, 1e6) * sqrt(nn * eps()) * norm(fjac2, 1)
+                    g = Ac_mul_B(fjac, fvec)
+                    p = (fjac2 + lambda * eye(nn)) \ g
+                    scale!(p, -1)
```
stevengj: What's going on here?  You are squaring the condition number by squaring the matrix.

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
...
+                else
+                    throw(e)
+                end
+            end
+            invfjac = inv(fjac)
```
stevengj: If you are computing `inv` anyway, you might as well compute it above.

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
...
+    oldinvfjac = alloc_jacobian(df, T, nn)
+    minvjΔf = similar(xold)
+    p = Array(T, nn)
+    g = Array(T, nn)
+    gr = Array(T, nn)
+
+    # Count function calls
+    f_calls::Int, g_calls::Int = 0, 0
+
+    df.fg!(x, fvec, fjac)
```
stevengj: One of the main attractions of the Broyden method is for when the Jacobian is not available.  Is there a way to use a finite-difference Jacobian as a fallback here?  It doesn't need to be super-accurate for Broyden to make progress.

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
...
+            end
+            invfjac = inv(fjac)
+            computejac = false
+        else
+            # the following corresponds to the Sherman-Morrison formula
+            # newJ^{-1} = J^{-1} + (Δx - J^{-1}Δf)Δx'J^{-1} / (Δx'J^{-1}Δf)
+            #TODO : Use QR update
+            Base.axpy!(-1.0, fvec, fvecold)
+            A_mul_B!(minvjΔf, invfjac, fvecold)
+            d = dot(p, minvjΔf)
```
stevengj: Where is `p` set?

src/broyden.jl
```diff
@@ -0,0 +1,170 @@
...
+            Base.axpy!(1.0, p, minvjΔf)
+            scale!(minvjΔf, - 1 / d)
+            A_mul_Bc!(scaleinvfjac, minvjΔf , p)
```
stevengj: This is a terribly expensive way to do a rank-1 update: you should never have to explicitly form the rank-1 matrix.  As it stands it seems like your complexity is O(n^3) (from the `A_mul_B!(invfjac, scaleinvfjac, oldinvfjac)`) rather than O(n^2).

src/broyden.jl
pkofod: This is how it used to be done in Optim, and I'm pretty sure NLsolve just mimicked it (we have gone away from this in Optim).

src/broyden.jl
pkofod: This is also due to code that was added to Optim a long time ago.

src/broyden.jl
pkofod: The decision to use analytical or approximated Jacobians is done at the time where `df` is constructed, so this line could very well be calling a finite difference version of the Jacobian.

pkofod:
@matthieugomez if you're up for it, let's see if we can address some of @stevengj 's comments here.

src/broyden.jl
ChrisRackauckas: Also, it should probably use `I` instead of `eye(nn)`

Issue #62, opened 2016-08-25
not_in_place + autodiff


matthieugomez:
I cannot use `not_in_place` and `autodiff` at the same time.

``` julia
 nlsolve(not_in_place(x-> x[1] + 2), [0.0], autodiff = true)
ERROR: unrecognized keyword argument "autodiff"
```

matthieugomez:
I'd also like to remove `not_in_place(f, g)` and `not_in_place(f, g, fg)`. 
I think it's neater if `not_in_place` and `n_ary` are just functionals, defined independent from the `DifferentiableFunction` type. 
Someone with a gradient could just do `not_in_place(f), not_in_place(g)`.  That being said, it removes the possibility to use `fg` not in place. But do people really use it?

KiranUofG:
Would it be possible to change the definition of not_in_place from not_in_place(f::function) to just not_in_place(f) ?

I use it for types, which are made callable by modifying the Base Call function. 

It works great when I make this change in my local NLsolve directory. While distributing my package however, I need to ask all my users to make this change in their local installations. 

I'm not sure if making this change would have implications for other applications. 

Please let me know if more info is required.

PR #66, opened 2016-09-04
Solve #62

matthieugomez:
Allow not_in_place / n_ary + autodiff.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -40,3 +40,27 @@ function check_isfinite(x::Vector)
         throw(IsFiniteException(i))
     end
 end
+
+
+function n_ary(f)
+    f!(x::Vector, fx::AbstractArray) = copy!(fx, [f(x...)... ])
+end
+
+function not_in_place(f)
+    f!(x::Vector, y::AbstractArray) = copy!(y, f(x))
+end
+
+function not_in_place(f, g)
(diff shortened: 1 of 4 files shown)


spencerlyon2:
Looks good to me. What do you think @KristofferC ?

src/utils.jl
```diff
@@ -39,4 +39,40 @@ function check_isfinite(x::Vector)
...
+function n_ary(f::Function)
+    f!(x::Vector, fx::AbstractArray) = copy!(fx, [f(x...)... ])
+end
+
+function not_in_place(f::Function)
+    f!(x::Vector, y::AbstractArray) = copy!(y, f(x))
+end
+
+function not_in_place(f::Function, g::Function)
+    depwarn("nlsolve(not_in_place(f, g)) is deprecated. Use nlsolve(not_in_place(f), not_in_place(g))")
```
KristofferC: Couldn't this function call what the deprecation is suggesting?

KristofferC:
Seems ok, no idea how many people use the not in place `fg` though. Probably not many.

src/utils.jl
matthieugomez: not exactly because one the replacement only works within nlsolve but I did change the function a little bit.

PR #67, opened 2016-09-04
Remove sumabs2j

matthieugomez:
sumabs2j is not needed anymore because view of sparse matrices is performant in 0.5


KristofferC:
Are they performant in 0.4 with Compat?

matthieugomez:
yes. It calls `slice` that works well too. I'm not sure when it changed.

PR #70, opened 2016-11-11
Fix isfinite depwarn

ChrisRackauckas:
`isfinite` was hit with a vectorization depwarn. I went and changed all instances to `isfinite.(x)`.

### DIFF
--- a/src/mcp.jl
+++ b/src/mcp.jl
@@ -16,10 +16,10 @@ function mcp_smooth(df::AbstractDifferentiableMultivariateFunction,
     function f!(x::Vector, fx::Vector)
         df.f!(x, fx)
         for i = 1:length(x)
-            if isfinite.(upper[i])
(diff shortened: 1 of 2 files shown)


spencerlyon2:
Thanks for doing this. I'd like to maintain 0.4 compatibility. Will you please use `@compat` in front of all these?

Thanks

ChrisRackauckas:
I added `@compat`

spencerlyon2:
Great, thanks.

Do we also need to bump the version of Compat in REQUIRE?

ChrisRackauckas:
I'm not quite sure how to find out what version it should be at. It should just need the broadcast parts.

spencerlyon2:
We're good on that. the broadcast notation hit Compat somewhere in the v0.7.x line -- we already require one of the v0.8.x releases.

PR #74, opened 2016-11-28
test on both 0.4 and 0.5 on travis

tkelman:
0.4 still supported according to REQUIRE


spencerlyon2:
Thanks, if it passes we'll merge

Issue #75, opened 2016-12-12
Autodifferentiation no longer works


ChrisRackauckas:
From the README

```julia
using NLsolve

function f!(x, fvec)
    fvec[1] = (x[1]+3)*(x[2]^3-7)+18
    fvec[2] = sin(x[2]*exp(x[1])-1)
end
nlsolve(f!, [ 0.1; 1.2],autodiff=true)
```

gives the error

```
MethodError: no method matching (::NLsolve.##26#29{#f!})(::Array{ForwardDiff.Dual{2,Float64},1})
Closest candidates are:
  #26(::Any, !Matched::Any) at /home/crackauc/.julia/v0.5/NLsolve/src/autodiff.jl:3
 in vector_mode_jacobian!(::DiffBase.DiffResult{1,Array{Float64,1},Tuple{Array{Float64,2}}}, ::NLsolve.##26#29{#f!}, ::Array{Float64,1}, ::ForwardDiff.JacobianConfig{2,Float64,Array{ForwardDiff.Dual{2,Float64},1}}) at /home/crackauc/.julia/v0.5/ForwardDiff/src/jacobian.jl:92
 in jacobian!(::DiffBase.DiffResult{1,Array{Float64,1},Tuple{Array{Float64,2}}}, ::NLsolve.##26#29{#f!}, ::Array{Float64,1}, ::ForwardDiff.JacobianConfig{2,Float64,Array{ForwardDiff.Dual{2,Float64},1}}) at /home/crackauc/.julia/v0.5/ForwardDiff/src/jacobian.jl:23
 in (::NLsolve.##28#31{NLsolve.##26#29{#f!}})(::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,2}) at /home/crackauc/.julia/v0.5/NLsolve/src/autodiff.jl:13
 in trust_region_(::NLsolve.DifferentiableMultivariateFunction, ::Array{Float64,1}, ::Float64, ::Float64, ::Int64, ::Bool, ::Bool, ::Bool, ::Float64, ::Bool) at /home/crackauc/.julia/v0.5/NLsolve/src/trust_region.jl:87
[...]

PR #76, opened 2016-12-12
fix AD and deprecations

KristofferC:
Had to comment out one test in minpack that started failing. Not sure why. Maybe because of changes to linesearches.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,5 +2,7 @@ julia 0.4
 Optim
 Calculus
 Distances
-ForwardDiff 0.2.0
+ForwardDiff 0.3.0
 Compat 0.8.4
+LineSearches
+DiffBase
--- a/src/NLsolve.jl
+++ b/src/NLsolve.jl
@@ -3,6 +3,7 @@ VERSION >= v"0.4.0-dev+6521" && __precompile__()
 module NLsolve
 
 using Distances
+using LineSearches
 using Optim
 using ForwardDiff
 using Compat
--- a/src/autodiff.jl
+++ b/src/autodiff.jl
@@ -3,15 +3,15 @@ function autodiff(f!, initial_x::Vector)
     permf! = (fx, x) -> f!(x, fx)
 
     fx2 = copy(initial_x)
-    g! = (x, gx) -> begin
-        out = ForwardDiff.JacobianResult(fx2, gx)
-        ForwardDiff.jacobian!(out, permf!, x)
-    end
+    jac_cfg = ForwardDiff.JacobianConfig(initial_x, initial_x)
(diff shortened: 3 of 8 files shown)


KristofferC:
Tests are passing locally. Not sure why they are failing on travis.

It is weird because the same test passes with AD but not with analytical derivative.
```jl
Brown almost-linear-newton                         40   515159    1000     2.050000e+01   6.757172e+00
minpack: Test Failed
  Expression: converged(r)
 in backtrace at error.jl:26
 in record at /home/travis/.julia/v0.4/BaseTestNext/src/BaseTestNext.jl:320
 in do_test at /home/travis/.julia/v0.4/BaseTestNext/src/BaseTestNext.jl:192
 [inlined code] from /home/travis/.julia/v0.4/NLsolve/test/minpack.jl:522
 in anonymous at no file:0
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:333
 [inlined code] from /home/travis/.julia/v0.4/NLsolve/test/runtests.jl:27
 in anonymous at no file:0
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:333
 in process_options at ./client.jl:280
 in _start at ./client.jl:378
Brown almost-linear-newton-AD                      40     461      18     1.137854e-09   3.268440e-02
```

On my computer:
```jl
Brown almost-linear-newton                         40     461      18     1.138134e-09   2.061997e-02
Brown almost-linear-newton-AD                      40     461      18     1.138135e-09   1.086409e-02
```

spencerlyon2:
Thanks for fixing this up.

That is strange. What do you recommend we do with the PR? Should we start travis again, or is it an issue that a simple restart won't fix?

KristofferC:
Since all the workers hit the same error I don't think a restart will do. However, I don't think I have introduced any bug with these changes or much more would fail. The only thing that could have changed is the backtracking linesearch has changed since the last time we ran the tests?

KristofferC:
Since all the workers hit the same error I don't think a restart will do. However, I don't think I have introduced any bug with these changes or much more would fail. But why would the same problem fail without AD but work with AD...

KristofferC:
Can we merge this even though the tests failed? I compared to earlier CI runs when tests passed and the results are identical for those minpack functions that converge. There is in my view almost zero probability that this PR introduces a bug that would cause only a few problems to fail.

I have high suspicion that the linea searches are the cause for this.

What do you think @spencerlyon2?

spencerlyon2:
I think it is Ok.  Let's merge

ChrisRackauckas:
Did this fix https://github.com/EconForge/NLsolve.jl/issues/75 ?

KristofferC:
Probably ;)

ChrisRackauckas:
Thanks. I suddenly had a test working again. Thanks for the New Year present!

Issue #78, opened 2016-12-21
v0.6 : tracing not defined


ChrisRackauckas:
Using the README example on master:

```julia
julia> using NLsolve

julia>

julia> function f!(x, fvec)
           fvec[1] = (x[1]+3)*(x[2]^3-7)+18
           fvec[2] = sin(x[2]*exp(x[1])-1)
       end
f! (generic function with 1 method)

julia>

julia> function g!(x, fjac)
           fjac[1, 1] = x[2]^3-7
           fjac[1, 2] = 3*x[2]^2*(x[1]+3)
           u = exp(x[1])*cos(x[2]*exp(x[1])-1)
           fjac[2, 1] = x[2]*u
           fjac[2, 2] = u
       end
g! (generic function with 1 method)

julia>

julia> nlsolve(f!, g!, [ 0.1; 1.2])
ERROR: UndefVarError: tracing not defined
 in trust_region_(::NLsolve.DifferentiableMultivariateFunction, ::Array{Float64,1}, ::Float64, ::Float64, ::Int64, ::Bool, ::Bool, ::Bool, ::Float64, ::Bool) at C:\Users\Chris\.julia\v0.6\NLsolve\src\trust_region.jl:3
 in #nlsolve#17(::Symbol, ::Float64, ::Float64, ::Int64, ::Bool, ::Bool, ::Bool, ::Function, ::Float64, ::Bool, ::NLsolve.#nlsolve, ::NLsolve.DifferentiableMultivariateFunction, ::Array{Float64,1}) at C:\Users\Chris\.julia\v0.6\NLsolve\src\nlsolve_func_defs.jl:24
 in (::NLsolve.#kw##nlsolve)(::Array{Any,1}, ::NLsolve.#nlsolve, ::NLsolve.DifferentiableMultivariateFunction, ::Array{Float64,1}) at .\<missing>:0
 in nlsolve(::Function, ::Function, ::Array{Float64,1}) at C:\Users\Chris\.julia\v0.6\NLsolve\src\nlsolve_func_defs.jl:45
```

PR #79, opened 2016-12-21
Fix macro hygine and sumabs2

ChrisRackauckas:
A macro hygine change on v0.6

https://github.com/JuliaLang/julia/pull/15850

breaks the macros used in this package. This is fixed by escaping the full quote, which should've been done in the first place (so it's a v0.5 and below compatible change).

Then, I also changed `sumabs2(x)` to `sum(abs2,x)` to fix the depwarns.

Now the example in the README no longer errors and no longer throws depwarns on v0.6.

Note this fixes https://github.com/EconForge/NLsolve.jl/issues/78


KristofferC:
LGTM.

PR #80, opened 2016-12-25
test travis with master

KristofferC:
just to see how tests are on travis with current code


KristofferC:
Failed on forward diff straight away so no info gotten...

PR #81, opened 2017-01-05
fix non initialized g

KristofferC:
`g` was used uninitialized in some cases. This lead to problems when `LineSearches` changed to use the old function value instead of recomputing a new one. Tests does not pass on latest LineSearches but it passes on https://github.com/anriseth/LineSearches.jl/pull/7


KristofferC:
This makes test pass on 0.4 and 0.5 at least (had to comment out some tests). I don't have time to dig deep down into why the linesearches fail, there have been large changes to them in LineSearches.jl. I will merge this and tag a new version because the current one doesn't even work properly.

PR #90, opened 2017-04-10
solve #62

matthieugomez:
Could we merge this?
Sorry for new pull request — I messed up the previous one #66 when trying to update it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,17 +176,16 @@ nlsolve(not_in_place(f), initial_x)
 Finite-differencing is used to compute the Jacobian in that case.
 
 If, in addition, there is a function `g(x::Vector)` returning a newly-allocated
-matrix containing the Jacobian, it can be passed as a second argument to
-`not_in_place`. Similarly, you can pass as a third argument a function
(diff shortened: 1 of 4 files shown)


KristofferC:
Isn't this removing the ability to use a `not_in_place` `fg`?

matthieugomez:
Ok, I have pushed a commit without depreciation messages for `not_in_place(f, g)` and `not_in_place(f, g, fg)`.  
I still think it would be better to depreciate them — I doubt people really use `not_in_place(f, g, fg)`.

sglyon:
I'm ok with the merge here. I think enabling Autodiff with `not_in_place` is more beneficial than preserving the `fg` version f `not_in_place`

KristofferC:
Sure, I just noted that the pull request did something extra than what #62 was about. I'm fine with removing the `fg` version of `not_in_place`.

matthieugomez:
Thanks everyone. Just to be clear the final commit does preserve the fg version of not_in_place.

PR #91, opened 2017-04-11
Change License

miguelraz:
Guys, it seems you forgot about this... 
https://github.com/JuliaNLSolvers/NLsolve.jl/issues/85
Got mentioned in chat, shoutout to gitter and @chrisrackauckas

If anything needs to be improved/got any comments critiques, they are more than welcome, as well as resubmissions.

I just used the wiki link to copy/paste it into here.
https://en.wikipedia.org/wiki/MIT_License


sglyon:
Thank you!

PR #95, opened 2017-04-24
Bugs

sglyon:
I've  just been going through and cleaning up a few bugs.

So far I have:

- In newton if the Jacobian is singular we were only catching `LAPACKException`, but should have also been looking out for `SingularException`.
- In trust region there were a couple things wrong:
    - the weighted norm function wasn't correct - it should have been applying the weights to both input vectors, but was only applying it once
    - The equation for `tau` didn't match Nocedal and Wright or what this package had just before this commit: 21fba867ef1c578f68971b79df4c2f7fba1d5c14. It is entirely possible that what is there right now is also correct, but I couldn't immediately see it so I just implemented the condition we had before that commit
    - I couldn't tell what was going on, so I added some comments

### DIFF
--- a/src/newton.jl
+++ b/src/newton.jl
@@ -76,7 +76,7 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
             df.f!(xlin, fvec)
             f_calls += 1
         end
-        return(dot(fvec, fvec) / 2)
+        dot(fvec, fvec) / 2
     end
 
     # The line search algorithm will want to first compute ∇fo(xₖ).
@@ -97,7 +97,7 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
     end
     function fgo!(xlin::Vector, storage::Vector)
         go!(xlin, storage)
-        return(dot(fvec, fvec) / 2)
+        dot(fvec, fvec) / 2
     end
 
     dfo = OnceDifferentiable(fo, go!, fgo!)
@@ -116,7 +116,7 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
             p = fjac\fvec
             scale!(p, -1)
         catch e
-            if isa(e, Base.LinAlg.LAPACKException)
+            if isa(e, Base.LinAlg.LAPACKException) || isa(e, Base.LinAlg.SingularException)
                 # Modify the search direction if the jacobian is singular
                 # FIXME: better selection for lambda, see Nocedal & Wright p. 289
                 fjac2 = fjac'*fjac
--- a/src/trust_region.jl
+++ b/src/trust_region.jl
@@ -20,7 +20,7 @@ macro trustregiontrace(stepnorm)
     end)
 end
 
-function dogleg!{T}(p::Vector{T}, r::Vector{T}, d2::Vector{T}, J::AbstractMatrix{T}, delta::Real)
+function dogleg!{T}(p::Vector{T}, r::Vector{T}, d::Vector{T}, J::AbstractMatrix{T}, delta::Real)
     local p_i
     try
         p_i = J \ r # Gauss-Newton step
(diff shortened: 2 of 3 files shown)


src/trust_region.jl
```diff
@@ -36,26 +36,44 @@ function dogleg!{T}(p::Vector{T}, r::Vector{T}, d2::Vector{T}, J::AbstractMatrix
...
+        # For intermediate we will use the output array `p` as a buffer to hold
+        # the gradient. To make it easy to remember which variable that array
+        # is representing we make g an alias to p and use g when we want the
+        # gradient
+
+        # compute g = J'r ./ (d .^ 2)
+        g = p
+        At_mul_B!(g, J, r)
+        broadcast!(/, g, g, d)
+        broadcast!(/, g, g, d)
```
KristofferC: Not needed in this PR but this can just be `g .= g./d`

KristofferC: Could fold them into one broadcast as well?

src/newton.jl
```diff
@@ -97,7 +97,7 @@ function newton_{T}(df::AbstractDifferentiableMultivariateFunction,
     end
     function fgo!(xlin::Vector, storage::Vector)
         go!(xlin, storage)
-        return(dot(fvec, fvec) / 2)
+        dot(fvec, fvec) / 2
```
KristofferC: Does this really help readability?

src/trust_region.jl
sglyon: Good call. I think we eventually just want `g .= g ./ (d .^ 2)`

I was doing it this way to avoid allocations on 0.5 where the dotted arithmetic operators don't get fused into the anonymous function handed to `broadcast!`

src/trust_region.jl
```diff
@@ -102,33 +120,34 @@ function trust_region_{T}(df::AbstractDifferentiableMultivariateFunction,
...
-
         it += 1
 
         # Compute proposed iteration step
-        dogleg!(p, r, d2, J, delta)
+        dogleg!(p, r, d, J, delta)
 
         copy!(xold, x)
-        Base.BLAS.axpy!(one(T), p, x)
+        Base.BLAS.axpy!(one(T), p, x)  # x += p
```
KristofferC: Could also be done with dot notation.

src/trust_region.jl
```diff
@@ -36,26 +36,44 @@ function dogleg!{T}(p::Vector{T}, r::Vector{T}, d2::Vector{T}, J::AbstractMatrix
...
+
             # Compute the optimal point on dogleg path
-            b = 2 * wdot(d2, p_c, p_diff)
-            a = wdot(d2, p_diff, p_diff)
-            c = wdot(d2, p_c, p_c)
-            tau = (-b + sqrt(b^2 - 4 * a * (c - delta^2)))/(2*a)
+            b = 2 * wdot(d, p_c, d, p_diff)
+            a = wnorm(d, p_diff)^2
+            tau = (-b+sqrt(b^2 - 4a*(wnorm(d, p_c)^2 - delta^2)))/(2a)
+            Base.BLAS.axpy!(tau, p_diff, p_c)  # p_c .+= tau .* p_diff
```
KristofferC: Dot notation

src/newton.jl
sglyon: This is just a personal style quirk. I don't like making return look like a function.

So I was going to change this to `return dot(fvec, fvec) / 2`. But then because I was making a change anyway I figured I might as well go all the way and adopt my preferred style of implicit return

sglyon:
@KristofferC thanks for the review on the dot notation changes.

I'm a bit torn here. Ultimately (e.g. post `v"0.6-"`) I think it is a great idea as it is much more readable and doesn't cost us much (just performance differences between some BLAS routines and native Julia ones). However, on 0.5 it results in the allocation of a number of temporaries.

What do people think?

src/utils.jl
```diff
@@ -1,18 +1,13 @@
-function wdot{T}(w::Vector{T}, x::Vector{T}, y::Vector{T})
+function wdot{T}(wx::AbstractVector{T}, x::AbstractVector{T},
+                 wy::AbstractVector{T}, y::AbstractVector{T})
     out = zero(T)
     @inbounds @simd for i in 1:length(x)
```
pkofod: I guess it won't really ever be, but if these aren't `Array`s, then it would be safer to go `for i in eachindex(wx, x, wy, y)`, but...

pkofod:
On the whole `v0.5` vs `v0.6`, I'm sort of torn. On one hand, it's kind of annoying for people who are invested in v0.5 to see performance degradations, on the other hand, `v0.6` is just around the corner. Actually, I'm maybe not that torn. In `Optim.jl`, I've generally just gone the "whatever is faster on `v0.6`" lately, because `v0.5` has a few bugs anyway, that is making `optimize` slower on `v0.5`. However, *when* v0.6 hits, I think there is no doubt: don't care about v0.5 if it isn't breaking.

src/newton.jl
KristofferC: I agree, remove the parenthesis. I just like explicit more than implicit. Don't care much though :)

KristofferC:
Without benchmarking, I would say these BLAS1 calls are not really performance critical compared to factorizing a whole matrix for example. So I would go with clarity.

sglyon:
Sounds good. 

How about for now we merge this as is, but at the same time open an issue to move to dot-notation calls once 0.6 has been released (... for long enough for me to work on it day to day and find time to make the adjustments here :wink:)?

Does that work with everyone?

KristofferC:
Good to me

sglyon:
As I already got the approval from @pkofod in his review I'll merge now.

PR #96, opened 2017-04-24
RFC: DFSANE algorithm

sglyon:
NOTE: This builds on top of #95 

In this PR I have taken a first stab at  implementing a derivative free method known as DFSANE. My implementation is largely based off the version in scipy, with some consultation of the papers below.

There are a few things that need to be addressed here before merging:

- Change to handling of keyword arguments in `nlsolve` methods. In this PR I chose to just splat them in all but one of the methods, and hand them off to the "main" method that calls the underlying solver. In this "main" method the common kwargs for all methods are type checked. The `df_sane` method allows for a few more parameters to be tuned, so right now they just get splatted even in this main method and handed on to a  `df_sane` function that does type checking.
- Addition of DataStructures as a dependency
- Inclusion of two non-monotone line search algorithms in the file df_sane.jl. Do these belong here, or are the potentially more generally applicable and candidates for moving to LineSearches.jl? cc @anriseth @pkofod 
- What do do about the scipy license? NLsolve.jl is MIT licensed. Scipy is BSD. I hope I'm ok in just making an attribution at the top of the file and copy/paste the scipy license, but I'm not sure on this one.
- More testing. 
    - I'm not directly testing the line search methods
    - I've just included df_sane as another algorithm choice in the minpack tests. It would be better to see where else it can succeed and try there.

### References

La Cruz, W. (2012). A projected derivative-free algorithm for nonlinear equations with convex constraints. Optimization Methods and Software, 29(1), 24–41. http://doi.org/10.1080/10556788.2012.721129

Cheng, W., & Li, D. H. (2009). A derivative-free nonmonotone line search and its application to the spectral residual method. IMA Journal of Numerical Analysis, 29(3), 814–825. http://doi.org/10.1093/imanum/drn019

Cruz, W. L., & Raydan, M. (2003). Nonmonotone Spectral Methods for Large-Scale Nonlinear Systems. Optimization Methods and Software, 18(5), 583–599. http://doi.org/10.1080/10556780310001610493

The file [scipy/optimize/_spectral.py](https://github.com/scipy/scipy/blob/2b7ce3919688d37f2620425ab6f87ad172f7425f/scipy/optimize/_spectral.py)

### DIFF
--- a/src/df_sane.jl
+++ b/src/df_sane.jl
@@ -1,0 +1,279 @@
+abstract NonMonLineSearch
+
+type ChengLineSearch{T} <: NonMonLineSearch
+    xp::Vector{T}
+    Fp::Vector{T}
+    Q::Float64
+    C::Float64
+    nx::Int
+    gamma::Float64
+    tau_min::Float64
+    tau_max::Float64
+    nu::Float64
+end
+
+_name(::Type{ChengLineSearch}) = "Cheng non-monotonic linesearch"
+
+function ChengLineSearch{T}(xp::Vector{T}, Fp::Vector{T}=similar(xp),
+                            n_prev::Int=10, Q::Float64=1.0, C::Float64=1.0,
+                            gamma::Float64=1e-4, tau_min::Float64=0.1,
+                            tau_max::Float64=0.5, nu::Float64=0.85)
+    nx = length(xp)
+    ChengLineSearch(xp, Fp, Q, C, nx, gamma, tau_min, tau_max, nu)
+end
+
+function (ls::ChengLineSearch)(func!, x_k, d, f_k, eta)
+    alpha_p = 1.0
+    alpha_m = 1.0
+    alpha = 1.0
+
+    fp = 1.0
+    nfev = 0
+    while true
+        for i in 1:ls.nx
+            ls.xp[i] = x_k[i] + alpha_p * d[i]
+        end
+        func!(ls.xp, ls.Fp); nfev += 1
+        fp = _fmerit(ls.Fp)
+
+        fp <= ls.C + eta - ls.gamma * alpha_p^2 * f_k && break
+
+        alpha_tp = alpha_p^2 * f_k / (fp + (2*alpha_p - 1)*f_k)
+
+        for i in 1:ls.nx
+            ls.xp[i] = x_k[i] - alpha_m * d[i]
+        end
+        func!(ls.xp, ls.Fp); nfev += 1
+        fp = _fmerit(ls.Fp)
+
+        fp <= ls.C + eta - ls.gamma * alpha_m^2 * f_k && break
+
+        alpha_tm = alpha_m^2 * f_k / (fp + (2*alpha_m - 1)*f_k)
+
+        alpha_p = clamp(alpha_tp, ls.tau_min * alpha_p, ls.tau_max * alpha_p)
+        alpha_m = clamp(alpha_tm, ls.tau_min * alpha_m, ls.tau_max * alpha_m)
+    end
+    # Update C and Q
+    Q_next = ls.nu * ls.Q + 1
+    ls.C = (ls.nu * ls.Q * (ls.C + eta) + fp) / Q_next
+    ls.Q = Q_next
+
+    return fp, nfev
+
+end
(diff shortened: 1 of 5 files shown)


pkofod:
We're going to need a proper (but simple ofc) description of what is in this pkg soon imo. Could you add a short intro to this method somewhere? Not like a derivation or convergence theory, just so a user or dev has an idea about what is going on. Could be a md file in a temp doc folder or something like that until we get docs up and running.

pkofod:
> Inclusion of two non-monotone line search algorithms in the file df_sane.jl. To these belong here, or are the potentially more generally applicable and candidates for moving to LineSearches.jl?

I havn't looked at the code yet, but unless there's something special about the code that really ties it to this method, then I say: by all means put it in `LineSearches.jl`!

pkofod:
We need a licensing expert to weigh in on that part, as I have no idea about these matters.

anriseth:
> Inclusion of two non-monotone line search algorithms in the file df_sane.jl. Do these belong here, or are the potentially more generally applicable and candidates for moving to LineSearches.jl?

It could probably go into LineSearches. Great to have more choices for people, as in my experience there is never one particular line search that is the *best*. Your implementations of linesearches as types are also a good push/inspiration for how we may implement https://github.com/JuliaNLSolvers/LineSearches.jl/issues/9 :smiley: 

Out of interest: Do they assume that the objective is of the form `f(x) = \|G(x)\|_2^2` for some vector field `G(x)`?

sglyon:
That's a really good question. I know that is what is used in this implementation, but I don't know if that is necessary. The last two papers I referred to should contain that information

pkofod:
> Out of interest: Do they assume that the objective is of the form f(x) = \|G(x)\|_2^2 for some vector field G(x)?

Yes, so this should probably be well-documented if we put them in LineSearches

pkofod:
What is DataStructures used for btw? I seem to be missing it.

sglyon:
I will definitely have to document things when I find the time. I opened this to see if anyone had any early comments, but it is not in good enough shape to consider merging.

I'm not exactly sure when I will get around to tidying this up. 

Looks like you've figured out why I used DataStructures

src/df_sane.jl
```diff
@@ -0,0 +1,318 @@
...
+                    σ_eps::Float64=1e-10,
+                    maxfev::Int=1_000,
+                    ftol::Float64=convert(T, 1e-10),
+                    xtol::Float64=1e-15,
+                    iterations::Int=3_000,
+                    store_trace::Bool=false,
+                    show_trace::Bool=false,
+                    extended_trace::Bool=false)
+
+    lsT = linesearch == :cruz ? CruzLineSearch : ChengLineSearch
```
pkofod: Why the symbols when you have a type to work off of?

sglyon: Good call, that should be changed.

anriseth:
LineSearches are now using the callable-types approach, so it should be quite straight-forward to merge your new linesearches (if we agreed that they belong there)

sglyon:
That's great, thanks for the heads up!

cortner:
Would love to have non-monotone linesearch in LineSearches!

pkofod:
@sglyon is it realistic that you'll come back and finish this at some point?

sglyon:
Haha yep, it is realistic -- though I can't promise exactly when.

I would appreciate some help moving the line searches over to LineSearches.jl. I would be able to clean up the NLsolve side of the algorithm after that. 

Could anyone help with the movement of the linesearch algorithms?

anriseth:
> Could anyone help with the movement of the linesearch algorithms?

We are looking at changing the LineSearches interface to do univariate optimization on an interval. It will hopefully be easier to move the linesearch algorithms after that is done.

Ref:
- https://github.com/JuliaNLSolvers/LineSearches.jl/issues/79#issuecomment-365389912 
- https://github.com/JuliaNLSolvers/Optim.jl/pull/543

pkofod:
Yes I will have a pr to be discussed soon

pkofod:
https://github.com/pkofod/NLSolvers.jl/blob/65034bb76cbf91e22784bd0156f28b85416adb2f/src/nlsolve/spectral/dfsane.jl has it if anyone needs it

PR #101, opened 2017-05-22
Basic implementation of Anderson acceleration

antoine-levitt:
This is in response to https://github.com/JuliaNLSolvers/NLsolve.jl/issues/88. This is a very basic implementation of Anderson acceleration but it works fine enough for well-conditioned problems.

A better implementation would use updates to qr factors, not implemented in julia yet (https://github.com/JuliaLang/julia/issues/20791). This would enable a better scaling with the history size, and more stability (through condition number monitoring).

### DIFF
--- a/src/anderson.jl
+++ b/src/anderson.jl
@@ -61,8 +61,7 @@
         new_x = copy(gs[:,1])
         if m_eff > 0
             mat[:, 1:m_eff] .= (gs[:,2:m_eff+1] .- xs[:,2:m_eff+1]) .- (gs[:,1] .- xs[:,1])
-            A_ldiv_B!(alphas[1:m_eff], mat[:,1:m_eff], xs[:,1] .- gs[:,1])
-            alphas[m_eff+1:end] .= zero(T)
+            alphas[1:m_eff] .= mat[:,1:m_eff] \ (xs[:,1] .- gs[:,1])
             for i = 1:m_eff
                 new_x .+= alphas[i].*(gs[:,i+1] .- gs[:,1])
             end
--- a/src/anderson.jl
+++ b/src/anderson.jl
@@ -16,6 +16,8 @@
     N = length(x0)
     xs = zeros(T,N,m+1) #ring buffer storing the iterates, from newest to oldest
     gs = zeros(T,N,m+1) #ring buffer storing the g of the iterates, from newest to oldest
+    residuals = zeros(T, N, m) #matrix of residuals used for the least-squares problem
+    alphas = zeros(T, m) #coefficients obtained by least-squares
     fx = similar(x0) # temp variable to store f!
     xs[:,1] = x0
     errs = zeros(iterations)
@@ -24,8 +26,6 @@
     tracing = store_trace || show_trace || extended_trace
     old_x = xs[:,1]
     x_converged, f_converged, converged = false, false, false
-    mat = zeros(T, N, m) #pre-allocation
-    alphas = zeros(T, m)
 
     for n = 1:iterations
         # fixed-point iteration
@@ -60,8 +60,8 @@
         m_eff = min(n-1,m)
         new_x = copy(gs[:,1])
         if m_eff > 0
-            mat[:, 1:m_eff] .= (gs[:,2:m_eff+1] .- xs[:,2:m_eff+1]) .- (gs[:,1] .- xs[:,1])
-            alphas[1:m_eff] .= mat[:,1:m_eff] \ (xs[:,1] .- gs[:,1])
+            residuals[:, 1:m_eff] .= (gs[:,2:m_eff+1] .- xs[:,2:m_eff+1]) .- (gs[:,1] .- xs[:,1])
+            alphas[1:m_eff] .= residuals[:,1:m_eff] \ (xs[:,1] .- gs[:,1])
             for i = 1:m_eff
                 new_x .+= alphas[i].*(gs[:,i+1] .- gs[:,1])
             end
@@ -73,8 +73,10 @@
         xs[:,1] = new_x
     end
 
+    # returning gs[:,1] rather than xs[:,1] would be better here if
+    # xn+1 = xn+beta*f(xn) is convergent, but the convergence
+    # criterion is not guaranteed
     return SolverResults("Anderson m=$m β=$β",
-                         # returning gs[:,1] would be slightly better here, but the fnorm is not guaranteed
                          x0, copy(xs[:,1]), maximum(abs,fx),
                          n, x_converged, xtol, f_converged, ftol, tr,
                          f_calls, 0)
--- a/src/anderson.jl
+++ b/src/anderson.jl
@@ -1,3 +1,5 @@
+# Notations from Walker & Ni, "Anderson acceleration for fixed-point iterations", SINUM 2011
+# Attempts to accelerate the iteration xn+1 = xn + β f(x)
 
 function anderson_{T}(df::AbstractDifferentiableMultivariateFunction,
                       x0::Vector{T},
--- a/src/anderson.jl
+++ b/src/anderson.jl
@@ -1,7 +1,7 @@
 # Notations from Walker & Ni, "Anderson acceleration for fixed-point iterations", SINUM 2011
 # Attempts to accelerate the iteration xn+1 = xn + β f(x)
 
-function anderson_{T}(df::AbstractDifferentiableMultivariateFunction,
+@views function anderson_{T}(df::AbstractDifferentiableMultivariateFunction,
                       x0::Vector{T},
                       xtol::T,
                       ftol::T,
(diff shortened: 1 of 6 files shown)


cortner:
nice - I look forward to trying this out.

sglyon:
Thanks for getting started on this!

The code is clean and easy to read 💯 

Have you tried running more of the example problems on this solver? Perhaps those in the minpack.jl test suite?

I'd be curious to see how it holds up compared to newton and trust region. 

I also think there will be some "easy" performance improvements on the table, but I'll let the algorithm settle and be well-tested before diving into them

antoine-levitt:
Anderson acceleration is not really a general-purpose routine for nonlinear equations, it's mostly useful when
1 the function to solve for is expensive
2 no jacobian is available
3 there's already a "good" underlying fixed-point iteration

NLsolve is a good place to put it because that would allow for an easy comparison with other methods of this kind (mostly Broyden, when that's implemented). It's harder to test for because it's not expected to work on generic nonlinear systems. OTOH in the linear regime it's basically GMRES, so when started near a solution and with a large enough history length it converges, which is what I tested for.

pkofod:
fwiw I think we should have a separate `fixedpoint` function for fixed point finders. Not necessarily here, and possibly in a separate package, but...

antoine-levitt:
Not sure about that: fixed-point iterations and nonlinear solvers are very similar, and having a unified interface allows to change solvers at will.

pkofod:
> Not sure about that: fixed-point iterations and nonlinear solvers are very similar, and having a unified interface allows to change solvers at will.

They are indeed very similar and finding a fixed point can be (and is) framed as solving a set of nonlinear equations. Solving nonlinear equations is not the same as finding a fixed point *in general* though, so I would prefer
```
# F(x) = x is solved by
fixedpoint(F, x0, ...)
# G(x) = 0 is solved by
solve(G, x0, ...)
```
Then of course, you can choose amongst all the nonlinear solvers in your fixedpoint call and it will automatically subtract `x`
```julia
fixedpoint(F, x0, Newton())
# is equivalent to
solve(x-> F(x)-x, x0, Newton())
```

pkofod:
To clarify: I'm not asking you to do that now, just saying that I think we should do it later.

src/anderson.jl
```diff
@@ -0,0 +1,92 @@
...
+    tr = SolverTrace()
+    tracing = store_trace || show_trace || extended_trace
+    old_x = xs[:,1]
+    x_converged, f_converged, converged = false, false, false
+
+    for n = 1:iterations
+        # fixed-point iteration
+        df.f!(xs[:,1], fx)
+        f_calls += 1
+        gs[:,1] .= xs[:,1] .+ β.*fx
```
ChrisRackauckas: Shouldn't views be used to cut down allocations?

src/anderson.jl
```diff
@@ -0,0 +1,92 @@
...
+        
+        if converged
+            break
+        end
+
+        #update of new_x
+        m_eff = min(n-1,m)
+        new_x = gs[:,1]
+        if m_eff > 0
+            mat = (gs[:,2:m_eff+1] .- xs[:,2:m_eff+1]) .- (gs[:,1] - xs[:,1])
```
ChrisRackauckas: Pre-allocate to get rid of the temporaries here and at 62?

ChrisRackauckas:
I think that for completeness it would be nice to have a simple Picard version as well, if not just for benchmarking and possibly identifying issues.

antoine-levitt:
Pre-allocation and views done, thanks! For simple Picard, just set m=0.

pkofod:
Tests seem to fail

pkofod:
Great! I'll have a closer look as soon as I find the time (hopefully tomorrow) so this doesn't end up sitting here over the summer.

src/anderson.jl
```diff
@@ -0,0 +1,94 @@
...
+    gs = zeros(T,N,m+1) #ring buffer storing the g of the iterates, from newest to oldest
+    fx = similar(x0) # temp variable to store f!
+    xs[:,1] = x0
+    errs = zeros(iterations)
+    n = 1
+    tr = SolverTrace()
+    tracing = store_trace || show_trace || extended_trace
+    old_x = xs[:,1]
+    x_converged, f_converged, converged = false, false, false
+    mat = zeros(T, N, m) #pre-allocation
```
pkofod: pre-allocation of what? could you simply state what it's used for?

src/anderson.jl
```diff
@@ -0,0 +1,94 @@
...
+
+function anderson{T}(df::AbstractDifferentiableMultivariateFunction,
+                     initial_x::Vector{T},
+                     xtol::Real,
+                     ftol::Real,
+                     iterations::Integer,
+                     store_trace::Bool,
+                     show_trace::Bool,
+                     extended_trace::Bool,
+                     hist_size::Integer,
```
pkofod: Why is this called `hist_size`? Should be `m`. Same all similar places below.

src/anderson.jl
```diff
@@ -0,0 +1,94 @@
+# Notations from Walker & Ni, "Anderson acceleration for fixed-point iterations", SINUM 2011
+# Attempts to accelerate the iteration xn+1 = xn + β f(x)
+
+@views function anderson_{T}(df::AbstractDifferentiableMultivariateFunction,
```
pkofod: Underscores for internal functions are usually placed in front

src/anderson.jl
```diff
@@ -0,0 +1,94 @@
...
+        end
+
+        xs = circshift(xs,(0,1)) # no in-place circshift, unfortunately...
+        gs = circshift(gs,(0,1))
+        old_x = m > 1 ? xs[:,2] : copy(xs[:,1])
+        xs[:,1] = new_x
+    end
+
+    return SolverResults("Anderson m=$m β=$β",
+                         # returning gs[:,1] would be slightly better here, but the fnorm is not guaranteed
```
pkofod: What do you mean by this comment?

src/anderson.jl
antoine-levitt: I just did the same as the other functions, e.g. newton_

src/anderson.jl
antoine-levitt: done, thanks

src/anderson.jl
antoine-levitt: This is the user-facing method, hist_size is more readable than m (but in the algorithm m is shorter and closer to the notations of Walker & Ni). Can change if you like.

src/anderson.jl
antoine-levitt: Usually when solving a nonlinear problem with anderson acceleration, you expect that the base iteration xn+1 = xn + beta f(xn) is convergent, but slowly, so that returning gs[:,1], which is xs[:,1]
 + beta xs[:,1], should be slightly better than xs[:,1]. The problem is that you don't know f() of this guy, so cannot guarantee that it satisfies the convergence criterion.

antoine-levitt:
f(x) = 0 <=> x = x + beta*f(x), so everything is a fixed-point. Of course, the FP iteration might not converge and it might not make sense. However, if you start it close to a zero of f and with a small beta, f(x) is almost affine f(x) = Ax-b, and Anderson acceleration is basically GMRES on Ax=b so should converge (with a large enough history size). This test checks this.

src/anderson.jl
pkofod: Alright, weird convention in NLsolve I guess. :+1: then

src/anderson.jl
pkofod: I think `m` is fine; that's what we use in LBFGS over at Optim, and I've never heard complaints. I guess it just has to be mentioned. Speaking of mentioned, you should probably add a note somewhere in the readme about this new method.

pkofod:
> f(x) = 0 <=> x = x + beta*f(x)

Right right, sorry. So to solve `F(x)=x` people should provide `f(x)=F(x)-x)`. This would be nice to make 100% clear in the readme entry. 

Personally I'd like to see a readme entry and hist_size -> m and then I think we're good to go!

test/2by2.jl
```diff
@@ -44,6 +44,12 @@ r = nlsolve(df, [ -0.5f0; 1.4f0], method = :newton, linesearch! = LineSearches.b
 @test converged(r)
 @test norm(r.zero - [ 0; 1]) < 1e-6
 
+# test local convergence of Anderson: close to a fixed-point and with
+# a small beta, f should be almost affine, in which case Anderson is
+# equivalent to GMRES and should converge
+r = nlsolve(df, [ 0.01; .99], method = :anderson, hist_size = 10, beta=.01)
```
pkofod: mmissed a hist_size

antoine-levitt: That'll teach me to push without testing... thanks!

pkofod:
Looks like NLsolve is ready to accept a new solver for the first time in quite some time! Thanks for playing along @antoine-levitt !

ChrisRackauckas:
Sweet!

antoine-levitt:
Great, thanks for the thorough review! (on this and other PRs)

PR #104, opened 2017-05-25
Deprecations and ForwardDiff fixes

KristofferC:


KristofferC:
I fought for a while but it's just not worth trying to support older version (0.5 is old by now...).

KristofferC:
@sglyon are you ok with moving to 0.6?

sglyon:
Ha I'm not sure 0.5 is old, but as long as there is a version that installs on 0.5 I'm ok with the latest releases targeting 0.6

PR #105, opened 2017-05-25
set minimum julia version to 0.6.0-pre

tkelman:
for older 0.6-dev versions that might not support struct or abstract type,
better to stick to the 0.5-compatible versions of the package


tkelman:
thanks

PR #106, opened 2017-05-25
use dot syntax

KristofferC:


src/trust_region.jl
```diff
@@ -143,18 +143,18 @@ function trust_region_{T}(df::AbstractDifferentiableMultivariateFunction,
...
 
         # Ratio of actual to predicted reduction (equation 11.47 in N&W)
         A_mul_B!(r_predict, J, p)
-        Base.BLAS.axpy!(one(T), r, r_predict)
+        r_predict .+= r
         rho = (sum(abs2,r) - sum(abs2,r_new)) / (sum(abs2,r) - sum(abs2,r_predict))
 
         if rho > eta
             # Successful iteration
-            copy!(r, r_new)
```
sglyon: Is this any better as a broadcast operation instead of just a copy?

KristofferC: Yeah, this could also be broadcasted.

KristofferC: It is the same performance. So depends what you mean with "better".

sglyon: I was thinking performance .

If they are the same then I'd id say the broadcast one is clearer.

Let's merge

KristofferC: the broadcast version of `scale!` is slower though

PR #107, opened 2017-05-30
Update NLsolve to NLSolversBase and new LineSearches

pkofod:
If we merge and tag this, I can fix Optim to not depend on ReverseDiff, fix the new ForwardDiff syntax, and tag a new version that is v0.6 only.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.6.0-pre
-Optim
 Calculus
 Distances
 ForwardDiff 0.5.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,3 +5,4 @@ Distances
 ForwardDiff 0.5.0
 LineSearches
 DiffBase
+NLSolversBase 2.1.0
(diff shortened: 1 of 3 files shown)


KristofferC:
Hmm, problem is that Pkg downgrades Optim to 0.5 to be able to upgrade ForwardDiff to 0.5 https://travis-ci.org/JuliaNLSolvers/NLsolve.jl/jobs/237406291#L192. So it seems that we need a new Optim tag with support for ForwardDiff 0.5 to be able to merge this?

pkofod:
I don't think so, but I think we may be able to remove Optim as a dependency right?

KristofferC:
Oh yeah, of course.

pkofod:
> Oh yeah, of course.

Which was kind of one of the points of having a SolversBase anyway.

KristofferC:
So plan here is to merge this so that Optim can be updated cleanly. Then update the NLsolve code to the latest linesearch syntax (and deprecate the keyword `linesearch!` for `linesearch`, see https://github.com/JuliaNLSolvers/Optim.jl/issues/285)

PR #117, opened 2017-08-28
Allow AbstractArray input

devmotion:
Just discovered https://github.com/JuliaNLSolvers/NLsolve.jl/issues/111 and tried to implement it. I added tests as well and adjusted the documentation.


src/anderson.jl
```diff
@@ -77,13 +77,13 @@
     # xn+1 = xn+beta*f(xn) is convergent, but the convergence
     # criterion is not guaranteed
     return SolverResults("Anderson m=$m β=$β",
-                         x0, copy(xs[:,1]), maximum(abs,fx),
+                         x0, reshape(xs[:,1], size(x0)...), maximum(abs,fx),
```
sglyon: Is there a reason we had a `copy` of `xs` before. This call to `reshape` will return a view into `xs` for Array (perhaps other AbstractArray).

I'm not claiming this is a problem, just asking for clarification on the semantics so we know this is ok to do.

src/autodiff.jl
```diff
@@ -1,16 +1,19 @@
-function autodiff(f!, initial_x::Vector)
+function autodiff(f!::Function, initial_x::AbstractArray)
```
sglyon: I think I prefer to remove the `::Function` restriction.

Sometimes I create a type to represent my objective function and then implement the `call` method on that type. 

This is really similar to a closure (_very_ similar to what Julia does under the hood for a closure -- maybe even exactly the same), but allows me to use the type elsewhere, which is convenient at times

src/autodiff.jl
```diff
@@ -1,16 +1,19 @@
...
 
-    permf! = (fx, x) -> f!(x, fx)
+    fvec! = reshape_f(f!, initial_x)
+    permf! = (fx::AbstractVector, x::AbstractVector) -> fvec!(x, fx)
 
-    fx2 = copy(initial_x)
-    jac_cfg = ForwardDiff.JacobianConfig(nothing, initial_x, initial_x)
-    g! = (x, gx) -> ForwardDiff.jacobian!(gx, permf!, fx2, x, jac_cfg)
+    fx2 = vec(copy(initial_x))
```
sglyon: What does `vec` do for  `AbstractArray` types other than `Array`? I realize this is implementation dependent, but is there a clear statement about _what_ it is supposed to do?

KristofferC:
Sorry, I will not be able to review this for a month because I am having many deadlines stacked on top of each other.

src/autodiff.jl
devmotion: The default implementation just reshapes arbitrary `AbstractArray` to a one-dimensional vector and returns the unchanged vector for `AbstractVector`, see https://github.com/JuliaLang/julia/blob/2b983d41f064c39c0b73f6c7695cc13e0de45c03/base/abstractarraymath.jl#L38-L39 with some additional documentation on top.

src/autodiff.jl
devmotion: Sure. Would you like me to remove these restrictions also in line
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/differentiable_functions.jl#L11
and at similar places?
Moreover, should it even be removed from fields in lines
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/differentiable_functions.jl#L4-L6
and at similar places, and be replaced with some parameters `F1`, `F2`, `F3`? I always thought one shouldn't use fields with abstract types...

src/anderson.jl
devmotion: No, it will return a view into a newly allocated vector `xs[:,1]`. The `copy` was needed there before to retrieve this newly allocated vector and not a  subarray - because of `@views` in line
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/anderson.jl#L4
`xs[:,1]` alone would have returned a view into the matrix `xs`. Since I moved `@views` to lines
https://github.com/devmotion/NLsolve.jl/blob/426c3c2698697f0a7c5054720706855aae9461b9/src/anderson.jl#L30
and
https://github.com/devmotion/NLsolve.jl/blob/426c3c2698697f0a7c5054720706855aae9461b9/src/anderson.jl#L27
(these are the only other places where it caused views), I could get rid of `copy` in the results and just had to reshape it to the dimensions of the input.

devmotion: By the way, for the same reason `copy` is used in lines
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/anderson.jl#L43-L44
and
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/anderson.jl#L61
and
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/anderson.jl#L72.

src/autodiff.jl
ChrisRackauckas: Yes, replace them with parameters and have the parameters `<:Any` (which is implicit BTW). Fields with abstract types will not be type stable when accessing. Inplace functions are kind of an oddity here though. The access is not type stable, so the return type isn't inferred. However, since there is no return that is actually used, this type instability doesn't propagate, so the real cost ends up being minimal. On the otherhand, if it's parameterized by the function, then each function you give is a different concrete subtype and thus requires another compilation, so there's a tradeoff there. Whether specializing on the function or not is a good idea depends on the use case. I think this is a good place to use FunctionWrappers.jl sometime in the future when that all stabilizes, but for now I would just stick with their choice here.

src/autodiff.jl
ChrisRackauckas: It's basically a view that changes the size so that way it will always index linearly.

devmotion:
I quite radically removed all `::Function` annotations, even from `linesearch!` keywords. Since `LineSearches` switched to callable types some time ago (see https://github.com/JuliaNLSolvers/LineSearches.jl/pull/43) this also enables the use of the new api of `LineSearches` and gets rid of deprecation warnings.

pkofod:
What I would love to see is a test using something that is an `AbstractArray` but not and `Array`.

devmotion:
I noticed that it's not a good idea to test `MappedArray` since it is copied to a regular `Array` by `copy`, and hence the tests won't catch any problems inside the algorithms. Thus I replaced the tests with a `DEDataArray` from `DiffEqBase` which is an `AbstractArray` type. A `DEDataArray` `A` wraps an array `A.x` and can contain additional fields. I had to remove/loosen some additional type annotations and also define `A_mul_B!` for this special type (the `reshape` method is just for convenience to ensure that the result of the algorithm is of the same type as the initial array). Tests currently fail since the Newton method needs an additional change in `NLSolversBase`, see https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/18.

ChrisRackauckas:
Re-run this test now?

devmotion:
It seems tests passed.

pkofod:
I'll have a final look tonight, and merge unless I find something I don't understand :)

README.md
```diff
@@ -70,9 +70,10 @@ Jacobian.
 
 This is the most efficient method, because it minimizes the memory allocations.
 
-In the following, it is assumed that have defined a function `f!(x::Vector,
-fx::Vector)` computing the residual of the system at point `x` and putting it
-into the `fx` argument.
+In the following, it is assumed that you have defined a function
+`f!(x::AbstractVector, fx::AbstractVector)` or, more general,
```
pkofod: "more generally"

src/autodiff.jl
```diff
@@ -1,16 +1,19 @@
-function autodiff(f!, initial_x::Vector)
+function autodiff(f!, initial_x::AbstractArray)
 
-    permf! = (fx, x) -> f!(x, fx)
+    fvec! = reshape_f(f!, initial_x)
+    permf! = (fx::AbstractVector, x::AbstractVector) -> fvec!(x, fx)
```
pkofod: Is there a specific reason why you're annotating this?

src/autodiff.jl
```diff
@@ -1,16 +1,19 @@
...
-    jac_cfg = ForwardDiff.JacobianConfig(nothing, initial_x, initial_x)
-    g! = (x, gx) -> ForwardDiff.jacobian!(gx, permf!, fx2, x, jac_cfg)
+    fx2 = vec(copy(initial_x))
+    jac_cfg = ForwardDiff.JacobianConfig(nothing, vec(initial_x), vec(initial_x))
+    function g!(x::AbstractVector, gx::AbstractMatrix)
+        ForwardDiff.jacobian!(gx, permf!, fx2, x, jac_cfg)
+    end
 
-    fg! = (x, fx, gx) -> begin
+    fg! = (x::AbstractVector, fx::AbstractVector, gx::AbstractMatrix) -> begin
```
pkofod: and here

src/differentiable_functions.jl
```diff
@@ -1,91 +1,159 @@
...
-struct DifferentiableMultivariateFunction <: AbstractDifferentiableMultivariateFunction
-    f!::Function
-    g!::Function
-    fg!::Function
+struct DifferentiableMultivariateFunction{F1,F2,F3} <: AbstractDifferentiableMultivariateFunction
+    f!::F1
+    g!::F2
+    fg!::F3
+
+    DifferentiableMultivariateFunction{F1,F2,F3}(f!, g!, fg!) where {F1,F2,F3} =
```
pkofod: why is this constructor needed?

src/differentiable_functions.jl
```diff
@@ -1,91 +1,159 @@
...
         f!(x, fx)
         g!(x, gx)
     end
     return DifferentiableMultivariateFunction(f!, g!, fg!)
 end
 
-function DifferentiableMultivariateFunction(f!::Function; dtype::Symbol=:central)
-    function fg!(x::Vector, fx::Vector, gx::Array)
+function DifferentiableMultivariateFunction(f!; dtype::Symbol=:central)
+    function fg!(x::AbstractVector, fx::AbstractVector, gx::AbstractMatrix)
```
pkofod: again, I'm wondering if this is really needed? I know the annotations were there, but...

src/differentiable_functions.jl
```diff
@@ -1,91 +1,159 @@
...
-struct DifferentiableSparseMultivariateFunction <: AbstractDifferentiableMultivariateFunction
-    f!::Function
-    g!::Function
-    fg!::Function
+struct DifferentiableSparseMultivariateFunction{F1,F2,F3} <: AbstractDifferentiableMultivariateFunction
+    f!::F1
+    g!::F2
+    fg!::F3
+
+    DifferentiableSparseMultivariateFunction{F1,F2,F3}(f!, g!, fg!) where {F1,F2,F3} =
```
pkofod: and here..? I know you probably added them for a reason, I'm just not sure why :)

src/differentiable_functions.jl
devmotion: No, it's not I guess though I haven't tried to remove the annotations. Most/all annotations are rather reminders and assertions to both show and ensure which type of function is created or expected. Do you want to get rid of annotations where possible?

src/differentiable_functions.jl
devmotion: The problem was that outer constructors such as `DifferentiableSparseMultivariateFunction(f!, g!, initial_x::AbstractArray)` were never called since these calls were already catched by the default constructors, and apparently it was solved by adding (slightly different) default constructors. However, the correct way is to just specify `DifferentiableSparseMultivariateFunction(f!::F1, g!::F2, initial_x::AbstractArray) where {F1,F2}`, I guess. I'll fix it.

pkofod:
Looks good to me, and tests pass, so... thanks!

ChrisRackauckas:
cheers!

pkofod:
Just waiting for tags

PR #129, opened 2017-11-18
Change to DiffEqDiffTools from Calculus.

pkofod:
To not allocate each time we need access to an `x`. That's not how it's currently done but it should come.

@dextorious any tips on calling this ? Should I add all the positional fields?


dextorious:
Generally, you should only add the positional fields that give some benefit to you. In this case, you'd need to add `Val{:Complex}` if you ever pass a complex-valued callable (real is the default and, for performance reasons, the code doesn't try to check). The rest of the positional arguments shouldn't matter for your use case (only forward differences make use of `fx`, since the API contract is that `fx` isn't modified by DiffEqDiffTools).

ChrisRackauckas:
0.2.0 looks like:

```julia
cache = DiffEqDiffTools.JacobianCache(similar(x),similar(x),similar(x))
DiffEqDiffTools.finite_difference_jacobian!(J,f, x, central_cache)
```

with sensical defaults and using the inplace function `f(out,x)`. Gradients still need some work so it's not ready for Optim, but the Jacobians I think are pretty much set (except for handling immutable inputs like StaticArrays, which I don't think NLsolve.jl handles anyways).

pkofod:
Good,  I like it. I'll fix it later and merge

pkofod:
I'll update it for the new syntax later (probably tomorrow) and merge.

PR #132, opened 2017-11-22
Get rid of prerelease

ChrisRackauckas:
Required, since it's causing https://github.com/JuliaLang/METADATA.jl/pull/12155 CI to fail.


pkofod:
Fixed!

PR #134, opened 2018-01-17
Update README.md

Sleort:
"[...] preallocated vector passed as second argument" --> "[...] preallocated vector passed as *first* argument"


pkofod:
Good catch! Thanks

PR #136, opened 2018-01-20
New inplace keyword to replace not_in_place functions.

pkofod:
This should take care of the last unfinished business on current master. This is another breaking change. I've purged the `not_in_place` functions for an `inplace` keyword to `nlsolve`. I think a keyword like this should probably be added in Optim as well. Maybe it should be implemented in `NDifferentiable` constructors directly as well, but that'll have to wait.

See the new README here https://github.com/JuliaNLSolvers/NLsolve.jl/blob/62a7da8aec95e808fecae1a2a4427d8c2b9f2545/README.md

Any comments are welcome. I'll merge soon though and tag a new version. Then we can take any confusion or subtle bugs as they appear. We really need to get up to date with NLSolversBase to not block the ecosystem, and the same goes for the positional argument order change.


matthieugomez:
It would be better to depreciate the function.

PR #137, opened 2018-01-24
Update autodiff.jl

pkofod:


ChrisRackauckas:
Tested this and it works well. Will this be released as v0.15.0 before the end of the week?

pkofod:
Perfect, I forgot to cc you. I'll release it later probably

PR #145, opened 2018-02-14
Update README.md

BeastyBlacksmith:
Added symbol for anderson acceleration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,9 @@ vector and evaluate the function at the new point.
 
 ## Anderson acceleration
 
-Also known as DIIS or Pulay mixing, this method is based on the
+This method is selected with `method = :anderson`.
+
+It is also known as DIIS or Pulay mixing, this method is based on the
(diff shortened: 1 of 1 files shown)


pkofod:
Thanks Beasty Blacksmith! Have you tried it out on a problem?

BeastyBlacksmith:
I tried, but it failed ( but it most probably wasn't suited for it ).

pkofod:
Did you play around with the memory length ? The default is 0 so that is effectively Picard iterations

BeastyBlacksmith:
I did, but my function tends to be constant a lot with "dips"

pkofod:
This is not a cross-examination, I'm just curious. What did you end up using?

BeastyBlacksmith:
At the moment I'm using `bboptimize` from BlackBoxOptim.jl, but I'm still completely satisfied. I mean I am doing some parameter estimation for boundary value problems on a set of ODEs.

PR #153, opened 2018-02-24
Broyden

pkofod:
Okay, this is just a quick prototype. It's not very robust, and it doesn't use LineSearches because I want that to be changed to the new API first (where you just pass in univariate functions).

@stevengj do you have any experience with "good" vs "bad" broyden? I seemed to get the idea from more recent discussions that "bad" broyden might actually have some merit. We should probably keep the Broyden type (see the struct in the top that is going to be used for a dispatch interface going forward) open to accept different specifications unless we "only" want "good" broyden.


pkofod:
Need to clean up some tests and allow NonDifferentiable, then it's done.

antoine-levitt:
Looks good! Out of curiosity, why implement this rather than a limited-memory version, ala LBFGS, maybe using https://github.com/timholy/WoodburyMatrices.jl? In my use cases, even forming a jacobian is unfeasible. Or would such a version be better suited as a variant of anderson?

pkofod:
I’ve been experimenting with a dedicated quasi-newton package “offline” and Broyden was one of the things I tried in that framework, so I decided to revive my old pr :) it was also a good reason to read up on derivative free line searches. I’d be interested in a large scale solvers as well!

pkofod:
I need to look into the line search parameters, but that's for a rainy day!

PR #156, opened 2018-03-31
change the way to reset the sparse matrix

KristofferC:
use official API, and also be a bit more clear that you often don't want to reset the sparsity pattern


pkofod:
Thanks

PR #171, opened 2018-06-22
 β to beta in README

CarloLucibello:
the README reads 
"this method is based on the acceleration of the fixed-point iteration `xn+1 = xn + β f(xn)`, where
by default `β=1`"
but I had to delve into the source code to figure out that the right keyword argument is `beta`.

Here I'm changing the keyword `beta` to `β`.
As an alternative we could change `β` to  `beta` in the README.

Cheers,
Carlo

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,15 +284,15 @@ vector and evaluate the function at the new point.
 This method is selected with `method = :anderson`.
 
 It is also known as DIIS or Pulay mixing, this method is based on the
-acceleration of the fixed-point iteration `xn+1 = xn + β f(xn)`, where
-by default `β=1`. It does not use Jacobian information or linesearch,
+acceleration of the fixed-point iteration `xₙ₊₁ = xₙ + beta*f(xₙ)`, where
+by default `beta=1`. It does not use Jacobian information or linesearch,
 but has a history whose size is controlled by the `m` parameter: `m=0`
 (the default) corresponds to the simple fixed-point iteration above,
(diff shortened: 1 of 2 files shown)


pkofod:
Thanks for the contribution! Clarity in documentation is very important.

I think it would be better to do it the other way around. I think Unicode is neat, but the general consensus is to not use these special characters in user facing code.

antoine-levitt:
Yes, thanks for tackling this overlook! Agreed, it's better not to use Unicode in external API functions. Alternatively, the Unicode can stay in the description of the doc, as long as the docstring of the function uses the plain text?

CarloLucibello:
there is no docstring for `nlsolve` (which is another issue by itself) and `anderson` (also not exported)

Should I change the readme to 
"this method is based on the acceleration of the fixed-point iteration `xn+1 = xn + beta*f(xn)`, where
by default `beta=1`"
?

pkofod:
> "this method is based on the acceleration of the fixed-point iteration xₙ₊₁ = xₙ + beta*f(xₙ), where
> by default beta=1"
> ?

Exactly, yes.

Doc strings and general code comments are very much valued as PRs! I plan to only focus on things that clarify the code (here and in Optim) in the immediate future (although the time I can spend is quite limited).

pkofod:
Thanks.

I was mainly worried about user facing APIs. Generally, it's OK that the field name is called unicode-beta as long as it is set in the constructor using the `beta` keyword.

PR #172, opened 2018-07-13
Fix some more Julia 0.7-beta deprecations

dawbarton:
This removes a few more Julia 0.7-beta deprecations. It changes `vecdot` -> `dot`, `vecnorm` -> `norm`, and `full` -> `convert`. It also adjusts the package tests to use `autodiff = :forward` where previously it was `autodiff = true`.

All tests now pass without deprecations.


pkofod:
Thanks a lot! (Sorry, I've been AFK for a few weeks)

Issue #152, opened 2018-02-23
Support Anderson Acceleration for fixedpoint


jlperla:
Right now the `:anderson` is implemented as a solution for solving nonlinear problems.  (where the solver turns the system into a fixed point problem.

Following something similar to https://github.com/JuliaNLSolvers/NLsolve.jl/issues/151 we could use the existing solver code by a small modification.

However, in the long-run it would make sense to rewrite (or copy) the Anderson code for direct fixed point problems to help its speed, as the above method would be taking a fixed point problem, converting it to a nonlinear system, and then converting it back to a fixed point problem in the underlying Anderson implementation.

ChrisRackauckas:
>However, in the long-run it would make sense to rewrite (or copy) the Anderson code for direct fixed point problems to help its speed

How would it help its speed? What part of the algorithm would be changed?

jlperla:
The tight inner loop in https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl would be a little different.  Longrun it is worth seeing it it matters.  Longrun.

ChrisRackauckas:
What would change in there?

antoine-levitt:
I'm not sure what you mean by this. Do you mean that there is the potential of inefficiencies because the code basically does x + (g(x) - x)? Sure. Anderson makes much more costly computations in its algorithm (and there are much bigger problems in the implementation, cf https://github.com/JuliaNLSolvers/NLsolve.jl/issues/128), so I'd be extremely surprised if that mattered. In any case I don't think it's worth complexifying the API/code.

jlperla:
Yes, that is what I meant.  I said longrun for a reason, I just didn't want to forget about it.

Looking at but very low priority, but this is the sort of algorithm to squeeze every imaginable efficiency out of (since it is often near the center of a whole bunch of type iterative procedures in economics algorithms).

Not worth spilling ink over, I just made a comment to point out that benchmarking would be worth looking at to see if it eliminates overhead in the implementation compared to embedded code.

ChrisRackauckas:
I would be very surprised if the difference it makes is greater than 0.1%

Issue #148, opened 2018-02-22
Implementation of "simple iteration" for the `fixedpoint`  with out of place functions.


jlperla:
I added in a shell for how out of place execution of the simple fixed point iteration.

1. In the committed files, you will see `/src/nlsolve/fixedpoint.jl`, which has the exported function for users to call.  I copied/pasted the `nlsolve.jl` code, so this may be imperfect.
2. That `fixedpoint(...)` function has placeholders to call various algorithms, which I commented out.  I put in a `/src/solves/simple_iteration.jl` file which we can hopefully use to get the basic fixed point iteration up and running.
3. I moved the code from our `test/fixed_points.jl` to this new  `/src/solves/simple_iteration.jl` and made a few tweaks.  You will see that I (1) removed the fixed point of a univariate function (which won't work the way nlsolve is defined); (2) left in code for the inplace `fixedpointOLD!` function, which is not implemented; and (3) changed the code for calling the function to use the `df.f()` notation (as this is the general way to wrap things in this library).

What I want to do is finish the implementation of `simple_iteration` for these fixed points.  Some of the obvious things to do:
- [ ] Read through an implementation of one of the solvers (e.g. newton) and see how it deals with tracing of results, etc.
- [ ] Right now it is just passing back a tuple of the result and number of iterations.  You should figure out what the `nlsolve` solvers return and the simple_iteration should return identical structures as the solver.
- [ ] Hook up any of the options passed into the `fixedpoint` function arguments to use in the `simple_iteration` algorithm.
- [ ] Finish up the (not inplace) tests
- [ ] I would see if you can mimic the `Cache` and `trace` options for this if possible (looking at something like the newton solver) but if it is difficult then it can be added later.

ArnavSood:
@AliKarimirad Here's a bit more granular take for how we can get started with this. (@jlperla let me know if it looks wrong to you). 

- [ ] Write a basic iteration scheme for a contraction mapping. The kind that most professors would write on the fly. 
- [ ] Extend that scheme (probably with different methods) to the different use cases for `NLsolve` (still without plugging into that caller). In particular, we need: 
               - In place vs. out of place functions. 
               - Functions that take several scalar arguments vs. an array (as per docs; just a question of uing the `n_ary` helper function).

This above piece will probably mainly involve lots of careful checking of method signatures to make sure everything is kosher, but the underlying math in each case should be the same. 

[...]

PR #174, opened 2018-07-23
Fixed point clean

arnavs:
@pkofod This PR contains the implementation of the `fixedpoint()` function talked about in e.g. #148 - #152. Specifically, here's what it contains: 

1. A wrapper `fixedpoint()` around `nlsolve()`, which takes the same type signature, applies one of two function maps (`f(x) -> f(x) - x`, and likewise for in-place), and forwards the result to `nlsolve()`. 

2. A set of tests for simple iteration (`:anderson` with `m = 0`), some more Anderson (`m = 1, 2, ...`), and some gradient methods (`:newton` with `autodiff = :central` and `autodiff = :forward`). 

3. Some brief notes in the README.md. 

Last I checked (yesterday), this is rebased with the lastest version of `master`, so the package tests as a whole pass, and Travis should be happy. 

cc: @jlperla @nosferican


README.md
```diff
@@ -314,6 +316,16 @@ Other optional arguments to `nlsolve`, available for all algorithms, are:
 * `extended_trace`: should additifonal algorithm internals be added to the state
   trace? Default: `false`.
 
+## Fixed Points 
+
+There is a `fixedpoint()` wrapper around `nlsolve()` which maps an input function `F(x)` to `G(x) = F(x) - x`, and likewise for the in-place. This allows convenient solution of fixed-point problems, e.g. of the kind commonly encountered in computational economics. Some notes:
+
+* The default method is `:anderson` with `m = 0`, which corresponds to simple iteration on a contraction mapping.
```
pkofod: It's clear that this will work on a contraction, but you could put in something that's not a contraction mapping, and hope for the best (not saying you should, but this is done in some cases).

Maybe something like: "The default method is `:anderson` with `m = 0`. This algorithm is robust when `F(x)` is a contraction mapping, but may not be very efficient when the Lipschitz constant is close to 1."

README.md
```diff
@@ -314,6 +316,16 @@ Other optional arguments to `nlsolve`, available for all algorithms, are:
...
+* Autodifferentiation is supported; e.g. `fixedpoint(f!, init_x; method = :newtonm. autodiff = :true`).
```
pkofod: I think this should say `... method = :newton, autodiff = :true)` right?
- `:newton` is misspelt `:newtonm` and a full stop is used instead of comma
- ) is outside the back ticks

README.md
arnavs: Right, you could do that and add that caveat. I've no issues with it. I just wanted to keep the docs as short as possible. But will amend the PR to make this clear.

README.md
```diff
@@ -314,6 +316,16 @@ Other optional arguments to `nlsolve`, available for all algorithms, are:
...
+* Tolerances and iteration bounds can be set exactly as in `nlsolve()`, since this function is a wrapper, e.g. `fixedpoint(f, init_x; inplace = false, iterations = 500, ...)`. 
+
+**Note:** If you are supplying your own derivative, it is currently not automatically mapped, so it must correspond with `F(x) = x`.
```
pkofod: "it is currently not automatically mapped, so it must correspond with `F(x) = x" is not super clear here, what does this actually mean, and how should the user work around this?

README.md
arnavs: Perfect, thanks. These are just errors.

REQUIRE
```diff
@@ -5,4 +5,4 @@ ForwardDiff 0.7.0
 LineSearches 5.0.0
 DiffBase
 NLSolversBase 6.0.0
-Reexport
+Reexport
```
pkofod: Why remove the newline char?

README.md
arnavs: The idea is that we currently don't map `J(x) -> J(x) - I` in the way that we map `f(x) -> f(x) - x`, in order to turn finding a 0 into finding a fixed-point. Will make that clearer.

REQUIRE
arnavs: This is something my editor does; I'm not sure why.

README.md
pkofod: Ah, okay, I get it. Why?

REQUIRE
pkofod: Sublime?

README.md
arnavs: It's on the list for things to implement. Some parts of this API are still in flux (like broadcasting on `UniformScaling`, so I'd rather wait until this is settled.

REQUIRE
arnavs: VS Code

README.md
arnavs: Done.

README.md
arnavs: Done.

README.md
arnavs: Done.

REQUIRE
arnavs: Added a newline here, I think.

src/objectives/autodiff.jl
```diff
@@ -0,0 +1,32 @@
+function OnceDifferentiable(f!, x::AbstractArray, F::AbstractArray, autodiff::Union{Symbol, Bool} = :central, chunk = ForwardDiff.Chunk(x))
```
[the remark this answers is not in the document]
arnavs: Deleted.

pkofod:
Will have a look

pkofod:
Maybe we should change the field name to `solution` instead as it is in either case the "solution" to the proposed problem -  or actually use a function based approach like in Optim. Not now though.

PR #179, opened 2018-07-31
Enable abstractarray tests

pkofod:
Not sure how @quinnj saw this, but I don't think this was meant to be disabled, or if it was then @arnavs will be able to tell us why :)


pkofod:
Seems fine, thanks so much for catching this @quinnj :) I think this might actually be a faulty rebase's fault, but who cares, it's fixed.

arnavs:
Wow, good catch. Yes, I think this was a rebasing artifact. Thanks for fixing :)

PR #183, opened 2018-08-13
Pass cashes through to inner function.

pkofod:
ref https://github.com/JuliaNLSolvers/NLsolve.jl/issues/181


pkofod:
cc @dawbarton  just so we don't both do it :)

PR #184, opened 2018-08-14
WIP: Initial attempt at an options structure for the Newton solver

dawbarton:
@pkofod as mentioned, here is my initial stab at the options structure based dispatch. I haven't hooked it into the `nlsolve` function yet and I'm still slightly hesitant as to whether to include the cache in the options (might you want the same options for multiple problems independent of the problem size?).

### DIFF
--- a/src/NLsolve.jl
+++ b/src/NLsolve.jl
@@ -31,7 +31,16 @@ export OnceDifferentiable,
        only_fj!,
        fixedpoint
 
+abstract type AbstractSolver end
 abstract type AbstractSolverCache end
+
+"""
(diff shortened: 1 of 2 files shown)


pkofod:
Are you familiar with `Optim.jl`? There we keep `linesearch` and `cache` out of the `Options` (that we just call `Options`, and instead have them in `Method` structs. I think we should do that here as well. By that I mean:

```
struct Newton
    linesearch
end
struct Broyden
    linesearch
end
struct Options
    tols...
    iterations
    trace stuff...
end
```
does that make sense? I'll be on github throughout the day tomorrow, and slack as well, so you can reach out.

PR #192, opened 2018-10-31
Change default `inplace` for `fixedpoint` to be `!applicable(f, initial_x)` instead of `true`

arnavs:
All tests pass.

I agree with @jlperla that this is a sensible change because (a) people aren't likely to define `f(out, x)` and `f(x)` using the same symbol (i.e., it would be `f!(out, x)`), and (b) it would save us the trouble of uglying code with `inplace = false` everywhere, when that kind of thing is inferrable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ similar to systems of nonlinear equations, except that the equality to zero is
 allowed to become an inequality if some boundary condition is satisfied. See
 further below for a formal definition and the related commands.
 
-There is also an identical API for solving fixed points (i.e., taking as input a function `F(x)`, and solving `F(x) = x`).
+There is also an almost-identical API for solving fixed points (i.e., taking as input a function `F(x)`, and solving `F(x) = x`).
 
 # Simple example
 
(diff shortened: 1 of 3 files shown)


pkofod:
I agree, I just didn’t get around to doing it the DiffEq way, but since there is (now?) an applicable function this is a no-brainer

pkofod:
You might want to change the `fixedpoint` tests to not have the `make_inplace`-thing while you're at it.

arnavs:
@pkofod I think it'd be best to push this change to students if it's good _per se_ (since we're using this `fixedpoint` function a lot, tagging dependency bundles, etc.). The tests can definitely be rewritten, but 'd rather do that in a dedicated pass.

pkofod:
> __ (since we're using this `fixedpoint` function a lot, tagging dependency bundles, etc.)

Alright... so is there a rush or isn't there a rush? :)

arnavs:
It would be good to get this done sooner rather than later. But it's not so urgent that you need to ram it through/stop people from weighing in.

arnavs:
Thanks @pkofod. Do you plan on tagging a new release soon?

pkofod:
Tomorrow, yes :)

Issue #187, opened 2018-09-19
Newton still uses `eye`


arnavs:
Get the `LoadError: UndefVarError: eye not defined` error when I use `method = :newton`. This is on the latest tagged release (`v2.1.0`). See below for some of the error message: 

```
Full Dynamic: Error During Test at /Users/arnavsood/.julia/dev/PerlaTonettiWaugh/test/runtests.jl:9
  Got exception outside of a @test
  LoadError: UndefVarError: eye not defined
  Stacktrace:
   [1] newton_(::OnceDifferentiable{Array{Float64,1},Array{Float64,2},Array{Float64,1}}, ::Array{Float64,1}, ::Float64, ::Float64, ::Int64, ::Bool, ::Bool, ::Bool, ::Static, ::NLsolve.NewtonCache{Array{Float64,1}}) at /Users/arnavsood/.julia/packages/NLsolve/2p0rf/src/solvers/newton.jl:108
   [2] newton_(::OnceDifferentiable{Array{Float64,1},Array{Float64,2},Array{Float64,1}}, ::Array{Float64,1}, ::Float64, ::Float64, ::Int64, ::Bool, ::Bool, ::Bool, ::Static) at /Users/arnavsood/.julia/packages/NLsolve/2p0rf/src/solvers/newton.jl:54
   [3] newton at /Users/arnavsood/.julia/packages/NLsolve/2p0rf/src/solvers/newton.jl:142 [inlined] (repeats 2 times)
   [4] #nlsolve#14(::Symbol, ::Float64, ::Float64, ::Int64, ::Bool, ::Bool, ::Bool, ::Static, ::Float64, ::Bool, ::Int64, ::Float64, ::typeof(nlsolve), ::OnceDifferentiable{Array{Float64,1},Array{Float64,2},Array{Float64,1}}, ::Array{Float64,1}) at /Users/arnavsood/.julia/packages/NLsolve/2p0rf/src/nlsolve/nlsolve.jl:20
[...]

PR #197, opened 2018-11-07
Fix the eye depreciation

matthieugomez:


pkofod:
Thanks; ref #187 I just never got around to it, sorry.

Is this a branch you're actually hitting? You may want to have a look at your problem :) (unless there's nothing to do; it's just if the Jacobian is actually singular at the solution, then you'll only get linear convergence, and Newton with line search might be overly expensive. Did you try the trust region one?)

pkofod:
https://github.com/JuliaLang/METADATA.jl/pull/19351

matthieugomez:
It’s just a very puzzling error message to get, so I wanted to remove it.

pkofod:
Should be gone now :) Thanks!

PR #200, opened 2018-12-08
broadcast for singular newton

matthieugomez:
The line failed because cache.p has the dimension of cache.g, not vec(cache.g). I have readded the test to make sure there is no runtime error, even though I do not assert convergence.


pkofod:
This is that singular newton again :) Thanks

Singularity is really a concern though, is it something you're hitting in each iteration or just occationally?

matthieugomez:
Just occasionally, and I'm not sure there is an actual solution in these cases.

pkofod:
> Just occasionally, and I'm not sure there is an actual solution in these cases.

Alright. Just wanted to be sure that you're aware that if the Jacobian is actually singular at the solution, Newton's method is potentially a very expensive way to achieve no more than linear convergence. We'd need to implement tensor methods for that, but it doesn't seem worth the trouble.

PR #203, opened 2018-12-23
Allow (and take advantage of) NonDifferentiable in Anderson.

pkofod:


pkofod:
I've had big plans for re-organizing this package for along time, but excuses, excuses and excuses. I really would want to branch off of traits on the algorithms instead of checking for `:anderson` (what if another acceleration scheme is added?), but this will work for now.

antoine-levitt:
> (what if another acceleration scheme is added?)

Acceleration I don't know, but there are a number of schemes (Broyden and its variants) that don't use jacobian information either.

PR #204, opened 2018-12-31
Fix printing in abstractarray tests and make newton's trace macro a f…

pkofod:
…unction call.


pkofod:
I cannot escape that weird `gmres!` thing in the linsolve tests... trying to print stuff out makes the error go away ...

PR #206, opened 2019-01-14
Passing a vector of f_tol's

BeastyBlacksmith:
This is a proposal to address https://github.com/JuliaNLSolvers/NLsolve.jl/issues/155


src/nlsolve/utils.jl
```diff
@@ -20,8 +20,14 @@ function assess_convergence(x,
         x_converged = true
     end
 
-    if maximum(abs, f) <= ftol
-        f_converged = true
+    if ftol isa AbstractArray
+        if all( f .< ftol )
```
antoine-levitt: Missing an `abs` there, and should be `.<=` for consistency. I think it's fine to keep this even for scalar `ftol`, even though it allocates more than `maximum(abs,f)<= tol`.

BeastyBlacksmith: Ok, I wasn't sure what the preferences are. I changed it accordingly

BeastyBlacksmith:
Actually this needs actualisation of the singatures of the solvers and some tests ( memo to myself ).

pkofod:
Stupid question: why not just scale your residuals in `f`?

pkofod:
To be quite explicit, let me explain why I'm a bit cautious: "stupid users". That might seem offensive, but it's not meant like that. I'm just curious that someone will look at their output and go: "oh, my third element doesn't converge.. I'll just set the f_tol to `[sqrt(eps(Float64)), sqrt(eps(Float64)), 1e-4, sqrt(eps(Float64)), sqrt(eps(Float64)), ...]` - you know? Now it "converges!, yay... but the users should probably have diagnosed why that third element isn't converging. Sure, it might just be a scaling issue, but in that case, why not just scale the residuals? 

Not against it, I'm just thinking about what we're buying ourselves here, and what bad behavior it might cause (I may be overdramatizing here, I'm just trying to debate it).

BeastyBlacksmith:
That also sounds like a good idea to me

PR #210, opened 2019-02-18
Use Walker's suggestions for Anderson acceleration

devmotion:
The current implementation of Anderson acceleration uses a different formulation of the least-squares problem than what was suggested by Walker and Ni in https://users.wpi.edu/~walker/Papers/Walker-Ni,SINUM,V49,1715-1735.pdf and https://users.wpi.edu/~walker/Papers/anderson_accn_algs_imps.pdf, and recomputes the QR decomposition in every step. In this PR I tried to follow those recommendations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,6 @@ julia> F
 * Broyden updating of Jacobian in trust-region
 * Homotopy methods
 * [LMMCP algorithm by C. Kanzow](http://www.mathematik.uni-wuerzburg.de/~kanzow/)
-* QR updating of the least-squares problem in the Anderson acceleration solver
 
 # Related Packages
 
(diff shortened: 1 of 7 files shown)


antoine-levitt:
Thanks, it looks great! @pkofod had a WIP version at https://github.com/pkofod/Anderson.jl, but this one seems more advanced. @pkofod?

antoine-levitt:
(fixes https://github.com/JuliaNLSolvers/NLsolve.jl/issues/128)

pkofod:
> Thanks, it looks great! @pkofod had a WIP version at https://github.com/pkofod/Anderson.jl, but this one seems more advanced. @pkofod?

Yeah, it looks good. I'm ashamed that it has been ignored for so long, but I've been busy elsewhere. I apologize.

Unless you've had any experience using this code since february @devmotion , and want to add something, I think this is pretty much good to go.

devmotion:
No worries, I didn't use or test the code in the last months but only added a similar implementation to OrdinaryDiffEq (that was specifically tailored to the use case there). So I'm fine with it :smiley:

PR #211, opened 2019-04-02
Remove DiffBase from REQUIRE

tkf:
DiffBase is not used anymore

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,7 +2,6 @@ julia 0.7.0-beta2
 Distances
 NLSolversBase 6.0.0
 LineSearches 5.0.0
-DiffBase
 DiffEqDiffTools
 ForwardDiff 0.7.0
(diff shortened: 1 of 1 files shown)


pkofod:
Seems like a pretty good argument :)

pkofod:
> DiffBase is not used anymore

how'd you stumble upon this btw?

tkf:
Thanks for the review and merge.

> > DiffBase is not used anymore
> 
> how'd you stumble upon this btw?

I created scripts to load the whole Manifest.toml and compile them into a system image https://github.com/tkf/julia-sysimage-recipes.  Then I realized that DiffBase was included due to NLsolve while it's source code is not Julia 1.0 compatible because it was giving me an import-time error.

pkofod:
Cool, thanks for making this PR as a result!

PR #222, opened 2019-08-12
Simplify nlsolve dispatches

devmotion:
This PR simplifies `nlsolve` by removing unneeded `where` clauses and forwarding keyword arguments.


pkofod:
Probably fine but have to see it on a computer. Thanks though :)

pkofod:
I forgot this... I wonder if all these keywords are tested :) But good idea to move the defaults to the inner methods.

PR #223, opened 2019-08-12
Fix Anderson for inputs of arbitrary dimension

devmotion:
This PR fixes the use of matrices (and higher-dimensional arrays) as inputs for Anderson.

The issue was not detected by the tests since `m = 0` was the default history size. I changed the defaults to `m = 10` and `droptol = 1e10`, as used in [Walker's implementation](https://users.wpi.edu/~walker/Papers/anderson_accn_algs_imps.pdf).

However, for the default setting `beta = 1` the Wood example diverges, leading to infinite values in the standard fixed-point iteration and `NaN` with Anderson acceleration (due to the computation of `Q` and `R`). Thus I explicitly chose `m = 2` and `beta = 1e-3`, such that the algorithm does not diverge within the first 1000 iterations, and added a check for finite values in the algorithm.

Moreover, I removed the `Anderson` struct which seemed unnecessary.

### DIFF
--- a/src/solvers/anderson.jl
+++ b/src/solvers/anderson.jl
@@ -45,7 +45,7 @@ function AndersonCache(df, m)
 end
 
 @views function anderson_(df::Union{NonDifferentiable, OnceDifferentiable},
-                             initial_x::AbstractArray{T},
+                             initial_x::AbstractArray,
                              xtol::Real,
                              ftol::Real,
                              iterations::Integer,
@@ -55,7 +55,7 @@ end
                              beta::Real,
                              aa_start::Integer,
                              droptol::Real,
-                             cache::AndersonCache) where T
+                             cache::AndersonCache)
     copyto!(cache.x, initial_x)
     tr = SolverTrace()
     tracing = store_trace || show_trace || extended_trace
@@ -89,7 +89,7 @@ end
             update!(tr,
                     iter,
                     maximum(abs, fx),
-                    iter > 1 ? sqeuclidean(cache.g, cache.x) : convert(real(T),NaN),
+                    iter > 1 ? sqeuclidean(cache.g, cache.x) : convert(real(eltype(initial_x)), NaN),
                     dt,
                     store_trace,
                     show_trace)
@@ -185,16 +185,16 @@ function anderson(df::Union{NonDifferentiable, OnceDifferentiable},
 end
 
 function anderson(df::Union{NonDifferentiable, OnceDifferentiable},
-                     initial_x::AbstractArray{T},
-                     xtol::Real,
-                     ftol::Real,
-                     iterations::Integer,
-                     store_trace::Bool,
-                     show_trace::Bool,
-                     extended_trace::Bool,
-                     beta::Real,
-                     aa_start::Integer,
-                     droptol::Real,
-                     cache::AndersonCache) where T
-    anderson_(df, initial_x, convert(real(T), xtol), convert(real(T), ftol), iterations, store_trace, show_trace, extended_trace, beta, aa_start, droptol, cache)
+                  initial_x::AbstractArray,
+                  xtol::Real,
+                  ftol::Real,
+                  iterations::Integer,
+                  store_trace::Bool,
+                  show_trace::Bool,
+                  extended_trace::Bool,
+                  beta::Real,
+                  aa_start::Integer,
+                  droptol::Real,
+                  cache::AndersonCache)
+    anderson_(df, initial_x, convert(real(eltype(initial_x)), xtol), convert(real(eltype(initial_x)), ftol), iterations, store_trace, show_trace, extended_trace, beta, aa_start, droptol, cache)
 end
--- a/src/solvers/anderson.jl
+++ b/src/solvers/anderson.jl
@@ -153,7 +153,7 @@ end
 
                 # solve least squares problem
                 γs = view(cache.γs, 1:m_eff)
-                ldiv!(R, mul!(γs, Q', fx))
+                ldiv!(R, mul!(γs, Q', vec(fx)))
 
                 # update next iterate
                 for i in 1:m_eff
--- a/src/solvers/anderson.jl
+++ b/src/solvers/anderson.jl
@@ -1,8 +1,6 @@
 # Notations from Walker & Ni, "Anderson acceleration for fixed-point iterations", SINUM 2011
 # Attempts to accelerate the iteration xₙ₊₁ = xₙ + beta*f(xₙ)
 
-struct Anderson{m} end
-
 struct AndersonCache{Tx,To,Tdg,Tg,TQ,TR} <: AbstractSolverCache
     x::Tx
     g::Tx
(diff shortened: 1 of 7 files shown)


src/solvers/anderson.jl
```diff
@@ -14,34 +12,40 @@ struct AndersonCache{Tx,To,Tdg,Tg,TQ,TR} <: AbstractSolverCache
...
+        Q = Matrix{T}(undef, length(x), mmax)
+        R = Matrix{T}(undef, mmax, mmax)
+    else
+        fxold = nothing
+        gold = nothing
+        Δgs = nothing
+        γs = nothing
+        Q = nothing
+        R = nothing
+    end
```
antoine-levitt: This is unstable. Not sure if that's a big problem in practice since small type unions are supposed to be fast these days, and anyway Anderson is mostly useful for large scale in which the overhead doesn't matter...

src/solvers/anderson.jl
```diff
@@ -14,34 +12,40 @@ struct AndersonCache{Tx,To,Tdg,Tg,TQ,TR} <: AbstractSolverCache
     R::TR
 end
 
-function AndersonCache(df, ::Anderson{m}) where m
+function AndersonCache(df, m)
```
antoine-levitt: I'm guessing this was done to have m be known to the compiler

devmotion: I think, the current implementation has the same type instability problem, it's just better hidden. In
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/ddd16a5675703cd2e3fd528f0dae5012a50255eb/src/solvers/anderson.jl#L175
the creation of `Anderson{m}` is also type unstable.

As far as I understand, both these instabilities in the current and the new implementation should not be a problem, since the type instability should not affect the output type when calling `nlsolve` and after the first call of `Anderson{m}` and `AndersonCache(df, m)` everything can be inferred. To me it seems, the current implementation is just more complicated without providing any benefits.

antoine-levitt: OK, I retract my objection then!

devmotion:
The following simple script (taken from `test/2by2.jl`)
```julia
using NLsolve
using InteractiveUtils

function f_2by2!(F, x)
    F[1] = (x[1]+3)*(x[2]^3-7)+18
    F[2] = sin(x[2]*exp(x[1])-1)
end

function g_2by2!(J, x)
    J[1, 1] = x[2]^3-7
    J[1, 2] = 3*x[2]^2*(x[1]+3)
    u = exp(x[1])*cos(x[2]*exp(x[1])-1)
    J[2, 1] = x[2]*u
    J[2, 2] = u
end

const df = OnceDifferentiable(f_2by2!, g_2by2!, [ -0.5; 1.4], [ -0.5; 1.4])

@code_warntype nlsolve(df, [ 0.01; .99], method = :anderson, m = 10, beta=.01)
```
shows that on Julia 1.1 the current implementation of `nlsolve` is actually type unstable
```julia
Body::NLsolve.SolverResults{Float64,Float64,Array{Float64,1},_1} where _1                  
```
while with this PR we get
```julia
Body::NLsolve.SolverResults{Float64,Float64,Array{Float64,1},Array{Float64,1}}
```
I'm not sure why the current implementation is type unstable since I don't know how to interpret the output of Cthulhu.

Comparing
```julia
f(df, m) = NLsolve.AndersonCache(df, Anderson{m}())

@code_warntype f(df, 0)
@code_warntype f(df, 10)
```
on master with
```julia
g(df, m) = NLsolve.AndersonCache(df, m)

@code_warntype g(df, 0)
@code_warntype g(df, 10)
```
on this PR shows that with the current implementation none of the type parameters can be inferred
```julia
Body::NLsolve.AndersonCache
1 ─ %1  = NLsolve.AndersonCache::Core.Compiler.Const(NLsolve.AndersonCache, false)
│   %2  = (Core.apply_type)(Main.Anderson, m)::Type
│   %3  = (%2)()::Any
│   %4  = (isa)(%3, NLsolve.Anderson{0})::Bool
└──       goto #3 if not %4
2 ─ %6  = (Base.getfield)(df, :x_f)::Array{Float64,1}
│   %7  = (Base.arraysize)(%6, 1)::Int64
│   %8  = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Float64,1}, svec(Any, Int64), :(:ccall), 2, Array{Float64,1}, :(%7), :(%7)))::Array{Float64,1}
│   %9  = (Base.getfield)(df, :x_f)::Array{Float64,1}
│   %10 = (Base.arraysize)(%9, 1)::Int64
│   %11 = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Float64,1}, svec(Any, Int64), :(:ccall), 2, Array{Float64,1}, :(%10), :(%10)))::Array{Float64,1}
│   %12 = %new(NLsolve.AndersonCache{Array{Float64,1},Nothing,Nothing,Nothing,Nothing,Nothing}, %8, %11, nothing, nothing, nothing, nothing, nothing, nothing)::NLsolve.AndersonCache{Array{Float64,1},Nothing,Nothing,Nothing,Nothing,Nothing}
└──       goto #4
3 ─ %14 = (%1)(df, %3)::NLsolve.AndersonCache
└──       goto #4
4 ┄ %16 = φ (#2 => %12, #3 => %14)::NLsolve.AndersonCache
└──       return %16
Body::NLsolve.AndersonCache
1 ─ %1  = NLsolve.AndersonCache::Core.Compiler.Const(NLsolve.AndersonCache, false)
│   %2  = (Core.apply_type)(Main.Anderson, m)::Type
│   %3  = (%2)()::Any
│   %4  = (isa)(%3, NLsolve.Anderson{0})::Bool
└──       goto #3 if not %4
2 ─ %6  = (Base.getfield)(df, :x_f)::Array{Float64,1}
│   %7  = (Base.arraysize)(%6, 1)::Int64
│   %8  = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Float64,1}, svec(Any, Int64), :(:ccall), 2, Array{Float64,1}, :(%7), :(%7)))::Array{Float64,1}
│   %9  = (Base.getfield)(df, :x_f)::Array{Float64,1}
│   %10 = (Base.arraysize)(%9, 1)::Int64
│   %11 = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Float64,1}, svec(Any, Int64), :(:ccall), 2, Array{Float64,1}, :(%10), :(%10)))::Array{Float64,1}
│   %12 = %new(NLsolve.AndersonCache{Array{Float64,1},Nothing,Nothing,Nothing,Nothing,Nothing}, %8, %11, nothing, nothing, nothing, nothing, nothing, nothing)::NLsolve.AndersonCache{Array{Float64,1},Nothing,Nothing,Nothing,Nothing,Nothing}
└──       goto #4
3 ─ %14 = (%1)(df, %3)::NLsolve.AndersonCache
└──       goto #4
4 ┄ %16 = φ (#2 => %12, #3 => %14)::NLsolve.AndersonCache
└──       return %16
```
whereas with this PR at least the first type parameter can be inferred:
```julia
Body::NLsolve.AndersonCache{Array{Float64,1},_1,_2,_3,_4,_5} where _5 where _4 where _3 where _2 where _1
1 ─ %1 = NLsolve.AndersonCache::Core.Compiler.Const(NLsolve.AndersonCache, false)
│   %2 = invoke %1(_2::OnceDifferentiable{Array{Float64,1},Array{Float64,2},Array{Float64,1}}, _3::Int64)::NLsolve.AndersonCache{Array{Float64,1},_1,_2,_3,_4,_5} where _5 where _4 where _3 where _2 where _1
└──      return %2
Body::NLsolve.AndersonCache{Array{Float64,1},_1,_2,_3,_4,_5} where _5 where _4 where _3 where _2 where _1
1 ─ %1 = NLsolve.AndersonCache::Core.Compiler.Const(NLsolve.AndersonCache, false)
│   %2 = invoke %1(_2::OnceDifferentiable{Array{Float64,1},Array{Float64,2},Array{Float64,1}}, _3::Int64)::NLsolve.AndersonCache{Array{Float64,1},_1,_2,_3,_4,_5} where _5 where _4 where _3 where _2 where _1
└──      return %2
```

So actually I think this PR improves type stability.

pkofod:
David,

Thanks! It looks good to me. If you agree it's done, then I'll merge.

devmotion:
Great! I don't want to add anything else to this PR :smiley:

pkofod:
Thanks!

PR #225, opened 2019-11-20
ged rid of dead code

KristofferC:


pkofod:
Yup, we're defaulting to Static from LineSearches now. Thanks

PR #255, opened 2020-11-10
Bump Distances.jl compat to 0.10

mateuszbaran:
Just bumping.

PR #257, opened 2020-11-19
Bump Distances.jl compat

dkarrasch:


mateuszbaran:
#255 does the same thing.

dkarrasch:
True.

PR #259, opened 2020-12-03
Changes to newton.jl so that newton with linesearches works with complex numbers

SarThak191119:
Hi,

I made some changes in NLsolve which did not earlier have a method to actually incorporate complex numbers into its newton algorithm using line searches. Linesearches.jl takes inherently real arguments. Nlsolve creates an objective single parameter function passed on to Linesearches which is defined by fo given by the dot product of the multivariate input function. If the input function is now complex the dot product still gives a real number but the dot operator returns a complex type with imaginary part zero. Taking the real part solves that issue and subsequently, all jacobian transpose has been changed to adjoint so that it is compatible with complex numbers.

The other change that has been done is with respect to the derivative of the univariate function which can be shown to be the real part of the product of cache.g and cache.p. 

Newton method with linesearches now converges to the roots of holomorphic multivariate functions. With non-holomorphic functions, it converges to roots if the initial starting point is sufficiently near.


pkofod:
Yes, this seems correct. This is the same we've done in Optim. Thanks a lot!

PR #271, opened 2021-06-15
Simpler first README example, also ref to Roots.jl.  Fixes #264.

### LINKED ISSUE #264: Redirect to Roots.jl for scrubs like myself
I came to this package something like separate 3 times, looking for a package to solve equations like exp(2x)=4. Simple univariate stuff. The first "simple example" shows me a bivariate function that mutates an input array to hold the output (New notation and use of a function to me), and defines a jacobian (which I have heard of but have no clue what is). Meeting a "simple example" that uses mathematical concepts that I don't know if was pretty intimidating, which turned me around in the doorway the first couple of times.

Considering the name of this package, I think it is quite natural that I ended up here. But it took quite the motivation and several attempts before first of all getting the guts to try my way on a "simple example", and secondly to realize that Roots.jl was what I was looking for.

Can something like the following be added to the start of the readme.md?

"For solving a single equation of a single variable like `exp(2x)+x^2=7`, a root-finding tool like Roots.jl is a more adequate/fitting tool." Possibly followed up with something like "Note that this can be accomplished by NLsolve as well, as shown below:
(insert the extra simple example)."

mauro3:
Whilst I've used a few bells and whistles of the excellent NLsolve, most of the time I just quickly want to solve a system of nl equations without any thoughts about performance.  This now shows this as the top-most example in the README.  Issue #264 suggests that at least one more person would appreciate a simpler first example (I also added the reference to Roots.jl).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,16 +22,39 @@ further below for a formal definition and the related commands.
 
 There is also an identical API for solving fixed points (i.e., taking as input a function `F(x)`, and solving `F(x) = x`).
 
-# Simple example
+Note, if a single equation and not a system is to be solved, consider using [Roots.jl](https://github.com/JuliaMath/Roots.jl).
+
+# A super simple example
 
 We consider the following bivariate function of two variables:
 
 ```jl
-(x, y) -> ((x+3)*(y^3-7)+18, sin(y*exp(x)-1))
+(x, y) -> [(x+3)*(y^3-7)+18, sin(y*exp(x)-1)]
 ```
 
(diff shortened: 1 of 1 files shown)


pkofod:
Thanks.

ChrisRackauckas:
Reference to NonlinearSolve.jl as a non-allocating version of Roots.jl?

mauro3:
Ok, I added that reference as a second commit.  As stated over in #264, a full summary of what package to use when would be good, but in another PR.

PR #274, opened 2021-08-13
Incorrect default in readme

jlperla:
In https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/nlsolve/nlsolve.jl#L14 it is `m=10` instead.


pkofod:
Thanks

PR #284, opened 2023-08-04
Add a naive project_region! option for Newton method

sjdaines:
Adds minimal option to modify Newton steps eg to project to region maintaining positive concentrations for chemical kinetics problems.

If using `method=:newton` with `linesearch=LineSearches.Static()` (the default linesearch), an additional argument `project_region!` (default value `(x)->(nothing)`) can be used to define a function to modify the proposed new `x` value after each Newton iteration.

For example, `project_region! = (x) -> (x .= max.(x, 1e-80))` will enforce a constraint `x[i] >= 1e-80`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,12 @@ By default, no linesearch is performed.
 **Note:** it is assumed that a passed linesearch function will at least update the solution
 vector and evaluate the function at the new point.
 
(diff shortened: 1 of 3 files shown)


nichollsh:
+1 for this - it would be fantastic!

PR #292, opened 2025-04-02
Create dependabot.yml and update actions

abhro:
From https://discourse.julialang.org/t/psa-use-dependabot-to-update-github-actions-automatically/96001


abhro:
Didn't realize renaming a branch closes the PR :sweat_smile:

PR #298, opened 2026-08-07
Fix CI so tests run again

pkofod:
CI has not run on this repo since #292 landed on 2 Dec 2025. This gets it running again, and fixes the first of two things that were broken underneath it.

## The workflow could not start

`CI.yml` referenced `julia-actions/setup-julia@2`. There is no such ref; the action publishes `v1`/`v2`/`v3`. GitHub could not resolve the action, so every job failed in under 10 seconds having executed zero steps. Both runs after #292 look like this:

```
test (1, ubuntu-latest)        failure   steps: []
test (1.0, ubuntu-latest)      failure   steps: []
test (nightly, ubuntu-latest)  cancelled steps: []
...
```

The workflow is now modelled on Optim.jl's:

- `setup-julia@v3`, `checkout@v6`, `julia-buildpkg@v1`, `julia-runtest@v1`, `codecov-action@v6`
- added `julia-actions/cache@v3`
- matrix is `lts` and `1` rather than `1.0` and `nightly`, and adds `windows-latest`
- `fail-fast: false`, so one bad cell no longer cancels the others
- coverage uploads from every cell, with `CODECOV_TOKEN`

`min` is deliberately absent, even though Optim.jl tests it. This package declares `julia = "1"`, so `min` resolves to 1.0.0 and would put CI straight back in the red. Adding it means bumping the compat bound to `"1.10"` in line with Optim/LineSearches/NLSolversBase, which is a support-policy decision and wants its own PR.

## The test suite could not load

`runtests.jl` defined `WrappedArray{T,N} <: DEDataArray{T,N}`, and `DEDataArray` no longer exists in DiffEqBase, so the suite raised `UndefVarError` before running a single test.

Every use of `WrappedArray` is commented out in `test/abstractarray.jl` (lines 37, 42, 45, 47, 50, 54, 57, 59, 62), so the type definition and its two `Base` methods were the only live references to it. Removed, along with `using DiffEqBase`, the `import Base.convert` that existed only to support it, and DiffEqBase from the test target. No coverage is lost. Restoring a custom-array test would be worth doing on its own terms, with a plain `AbstractArray` subtype.

## A real bug the dead CI was hiding

With the suite loading again, `test/complex.jl` turned up a genuine bug in `newton.jl`. `fo` computes the merit function as `real(dot(value(df), value(df))) / 2`, but `fgo!` computed `dot(value(df), value(df)) / 2` with no `real()`. For a complex-valued `f`, `dot` returns a `Complex` with a tiny imaginary part, and the linesearch objective stores `F` as `zero(real(T))`, so the assignment threw:

```
InexactError: Float64(34.52505 + 3.5438318946035e-16im)
```

Every non-`Static` linesearch was therefore unusable on complex problems. All four threw on the first iteration; all four converge with `real()` added:

| linesearch | before | after |
| --- | --- | --- |
| `BackTracking` | `InexactError` | converged, resnorm 2.4e-11 |
| `StrongWolfe` | `InexactError` | converged, resnorm 1.3e-12 |
| `HagerZhang` | `InexactError` | converged, resnorm 4.9e-15 |
| `MoreThuente` | `InexactError` | converged, resnorm 8.0e-11 |

`test/complex.jl` exercises exactly this and never caught it, because the DI error below aborted the testset before reaching that loop.

## Marked broken: the real-embedded comparison

`test/complex.jl` also errored with a `PreparationMismatchError` from DifferentiationInterface, which is what aborted the testset and took `interface/caches.jl` and `fixedpoint/fixedpoint.jl` down with it, since neither had been running at all. That one is upstream and is marked `@test_broken` here rather than fixed.

It reproduces with no NLsolve involved, on NLSolversBase 7.10.0 / Julia 1.10.11:

```julia
using NLSolversBase
f!(F, x) = (F[1] = x[1]^2 - 2; F[2] = x[2]^2 - 3)

x0 = reinterpret(Float64, [1.0+1.0im])
typeof(x0)          # Base.ReinterpretArray{Float64, 1, ComplexF64, Vector{ComplexF64}, false}

df = OnceDifferentiable(f!, x0, copy(x0); autodiff=:central)
typeof(df.x_f)      # Vector{Float64}

value_jacobian!!(df, copy(df.x_f))   # PreparationMismatchError
```

```
PreparationMismatchError (inconsistent types between preparation and execution):
  - f!: ✅
  - y: ✅
  - backend: ✅
  - x: ❌
    - prep: Base.ReinterpretArray{Float64, 1, ComplexF64, Vector{ComplexF64}, false}
    - exec: Vector{Float64}
  - contexts: ✅
```

The finite-difference branch of the `OnceDifferentiable` constructor calls `DI.prepare_jacobian(f, F2, backend, x_seed)` at `src/objective_types/oncedifferentiable.jl:83`, seeding preparation with the array the caller passed in. The object then stores `x_f` and `x_df` as `copy(x)`, and `copy` of a `ReinterpretArray` returns a plain `Vector`. Every subsequent evaluation hands DI a type preparation never saw, and DI's strict check rejects it.

Any wrapper array whose `copy` narrows to `Array` hits this. It surfaces here because comparing against the real-embedded formulation passes `reinterpret(Float64, ::Vector{ComplexF64})` as `initial_x`. This package's compat is `NLSolversBase = "7"`, so it arrived through a compatible dependency upgrade with no change on the NLsolve side.

The fix presumably belongs upstream: seed preparation with the array that will actually be used, `x_f`, so preparation and execution agree by construction. `strict=Val(false)` would silence the check but leaves preparation specialised on a type that never appears at execution.

Only the methods that build a `OnceDifferentiable` are affected. `:anderson` builds a `NonDifferentiable`, prepares no jacobian, and compares fine, so it stays a live `@test`. The complex solves themselves are now asserted directly rather than only relative to the real-embedded run, so that side keeps its coverage. `@test_broken` errors on an unexpected pass, so this reverts to `@test` on its own once the upstream fix lands.

## Result

Suite is green. `complex` reports 15 pass, 6 broken; `singular case` keeps its 2 intentional broken tests. `interface/caches.jl` and `fixedpoint/fixedpoint.jl` run for the first time in a while and pass.

---

This pull request was written with the assistance of generative AI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,29 +1,34 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - master
-    tags: '*'
-  pull_request:
+  - push
+  - pull_request
 
 jobs:
   test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        julia-version: ['1.0', '1', 'nightly']
-        os: [ubuntu-latest, macOS-latest]
-
+        version:
+          - "lts"
+          - "1"
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
     steps:
-      - uses: actions/checkout@v5
-      - uses: julia-actions/setup-julia@2
+      - uses: actions/checkout@v6
+      - uses: julia-actions/setup-julia@v3
         with:
-          version: ${{ matrix.julia-version }}
-      - uses: julia-actions/julia-buildpkg@latest
-      - uses: julia-actions/julia-runtest@latest
+          version: ${{ matrix.version }}
+      - uses: julia-actions/cache@v3
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
-      - uses: codecov/codecov-action@v5
-        if: ${{ matrix.julia-version == '1' && matrix.os == 'ubuntu-latest' }}
+      - uses: codecov/codecov-action@v6
         with:
+          fail_ci_if_error: true
           files: lcov.info
+          token: ${{ secrets.CODECOV_TOKEN }}
--- a/Project.toml
+++ b/Project.toml
@@ -18,11 +18,10 @@ Reexport = "0.2, 1.0"
 julia = "1"
 
 [extras]
-DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["DiffEqBase", "IterativeSolvers", "Random", "SparseArrays", "Test"]
+test = ["IterativeSolvers", "Random", "SparseArrays", "Test"]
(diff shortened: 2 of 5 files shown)


pkofod:
Filed the upstream half as JuliaNLSolvers/NLSolversBase.jl#172. It reproduces on NLSolversBase master (v8.0.0) as well as the released v7.10.0, and affects the gradient path as well as the jacobian path, so it is not specific to this package's usage.

The `@test_broken` markers here revert to `@test` once that lands.

